PR #1, opened 2018-11-20
Initial publication list

jppelteret:
Add the initial lists of publications (as scraped directly off the website's current publication.html).

### DIFF
--- a/publications-2007.bib
+++ b/publications-2007.bib
@@ -1,0 +1,408 @@
+% Encoding: ISO-8859-1
+
+@article{2007_0,
+   author  = "N. Antonic and M. Vrdoljak",
+   title   = "Gradient methods for multiple state optimal design problems",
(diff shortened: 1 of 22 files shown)


publications-2007.bib
```diff
@@ -0,0 +1,408 @@
...
+   title   = "Reconstruction of absorption and fluorescence contrast for scanning time-domain fluorescence mammography",
+   journal = "Optical Tomography and Spectroscopy of Tissue VII, Proc. SPIE 64340H/1-11",
+   year    = "2007",
+   volume  = "6434",
+   pages   = "",
+   url     = "",
+   doi     = ""
+}
+
+@phdthesis{2007_64,
```
jppelteret: @bangerth In the original `publications.bib` that you gave me, this entry was among some listed for 2013. I think that's why its key index is 64, while the previous entry's is only 37.

PR #2, opened 2018-11-20
Add new entries accumulated offline

jppelteret:
These are the new entries that @bangerth had accumulated offline. All "removed" entries were those that had a duplicate in the offline list. These have been updated and replaced by the new entries.


bangerth:
This seems like far more than I had in the file I sent to you and @tjhei a couple of weeks ago. Or are these the ones that @tjhei had sent around during the workshop this summer?

jppelteret:
This is the list that Timo sent (see email 26 July 2018 @ 17:41).

According to him,
> This contains 168 entries (the raw import from google scholar for 2016 was 240).

But I see now that you also sent a file `new-references.txt` afterwards, which I missed. So let me add those entries to and double check that the entries added in this PR do in fact cite us (something that I forgot to double check for all of them, doh!). I'll ping you when this is done.

tjhei:
> I'll ping you when this is done.

Why not a separate PR for those?

jppelteret:
> double check that the entries added in this PR do in fact cite us

I guess either way I should still do this step.

bangerth:
Let's do it in a separate PR. I have a couple more since then, so let's not hold up the process.

All the ones in my files have already been checked that they use deal.II.

jppelteret:
Ok. I'm working through cleaning and fixing up the already committed entries...

bangerth:
Oh, my apologies. It seemed implausible to me that you're working on this at this time of day :-( But I should have just looked at the time of your emails.

Since you have the current files on your drive, let me revert this. Then save your version, go to mainline, and you create a new patch on top of it?

jppelteret:
I've re-added most of these entries now (pushed straight to master). Some did not appear to cite the library so they were removed from the list. The others that I can't confirm either way (e.g. paywalled) I've left in there with a note that should hopefully remind me to check them later.

All of the entries render *something*, but I'm yet to scroll through all 89 pages of rendered PDF to check the correctness. I think that I've done as much as I can for now -- I guess its up to @bangerth and @tjhei to move forward from here.

PR #3, opened 2018-11-20
Add html generation

tjhei:
- simple version stolen from ASPECT
- uses the jabref inside a docker container
- builds ``publications.html``

We probably need to edit/replace the layout, but this at least "works".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+
+publications.html: Makefile
+	rm -rf temp/ publications.html
+	mkdir temp
+	cat publications-*.bib >temp/aspect.bib
+	cp jabref-template/* temp/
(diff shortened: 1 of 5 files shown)


tjhei:
umlauts are broken, see "Comparison of implicit and explicit". Let me play with this locally some more.

jppelteret:
I'm busy running these through latex (offline), and they're broken there. I guess I should fix those up first. I'll open a PR now for a latex project that builds all of the entries into a PDF.

tjhei:
okay, want to merge this as is for now and see if our other changes fixes the umlauts?

jppelteret:
Sure, sounds good!

PR #4, opened 2018-11-20
Add latex project to render the bibtex entries offine.

jppelteret:
This can be built using the command "latexmk publication_list.tex" and
can be used to check that the entries are correctly parsed and emit no
errors.


tjhei:
> @tjhei Maybe we could, at a later stage, add this to some sort of CI build to check that no errors are emitted when a new entry is added?

Yes. We will do that. Let's get it running first, though.

PR #5, opened 2018-11-20
add automated testing with Jenkins

tjhei:


jppelteret:
Cool :-)

jppelteret:
> ! LaTeX Error: File `biblatex.sty' not found.

Drats, do we need to add all of the package style files to the `offline` folder? (I see that you've already done this with `etoolbox.sty`.) If so, then that'd be annoying... 

I have one concern about doing this. I think that this would override the equivalent files in an existing tex distribution, and may break building on a local machine (e.g. if they are incompatible with newer/older tex versions). Should / Could we perhaps put these files in a folder `offline/jenkins` and copy them to the `offline` folder when setting up a build on Jenkins?

tjhei:
> With much fiddling I think I've got the LaTeX compilation working :-)

nooo. Sorry, I think you wasted your time here. I was just going to update latex in the container to include more packages.

jppelteret:
Ha ha, no worries. I would normally have asked first, but with the time difference and all I thought I I'd just get on with it. Either way, it was still an interesting exercise :-)

tjhei:
okay, this is now working.

tjhei:
(and the pdf and html file can be downloaded from the "artifacts" tab)

jppelteret:
So I had a look at the rendered PDF, and the numbering is a bit funny. Do we care about this?

jppelteret:
I checked the output on my machine and, with the build steps you use here, I get a PDF with the correct numbering. A clean build environment is used each time, right? I find that I have to delete all of the auxiliary files between builds to ensure that biber correctly.

tjhei:
Yes, the env is always clean. This change should work.

tjhei:
I think we should merge as is and improve later if needed.

jppelteret:
> I think we should merge as is and improve later if needed.

Agreed. It seems to be working as expected now. Thanks a lot!

> Yes, the env is always clean. 

This is was the only sensible answer to my dumb question. Its strange that it didn't work with the double compilation after building the bib index.

PR #7, opened 2018-11-20
Clean up original entries

jppelteret:
Changes Jabref entries to use brackets, guarding against runaway latex commands. Removes unicode and HTML from all entries.


jppelteret:
@bangerth Here you go. @tjhei I think that after this the CI machinery should work.

PR #10, opened 2018-11-21
Change encoding enforced by Jabref.

jppelteret:
The encoding is changed from ISO-8859-1 ( https://en.wikipedia.org/wiki/ISO/IEC_8859-1 ), which supports some special characters, to ASCII ( https://en.wikipedia.org/wiki/ASCII ), which does not. The intent is that Jabref will not allow files with special characters to be saved if the encoding does not allow it. This should help with detection if some anomalous characters slip into an entry (e.g. if entries are downloaded from DOI, e-print, etc.).

I understand that this is a bit hostile towards our European colleagues, but I think it will help ensure that all entries are rendered as intended. All special characters will thus need to be explicitly formed with tex commands. If it makes a difference, we already enforce this when `offline/publication_list.tex` is built.

### DIFF
--- a/publications-2014.bib
+++ b/publications-2014.bib
@@ -1,4 +1,4 @@
-% Encoding: ISO-8859-1
+% Encoding: US-ASCII
 
 @article{2014_0,
    author  = {S. G. Abaimov and J. P. Cusumano},
--- a/publications-2014.bib
+++ b/publications-2014.bib
@@ -1175,7 +1175,7 @@
 }
 
 @article{2014_110,
-   author  = {A. Vidal-Ferrandiza and R. Fayeza and D. Ginestarb and G. Verdúa},
+   author  = {A. Vidal-Ferrandiz and R. Fayez and D. Ginestar and G. Verd\'{u}},
    title   = {Solution of the Lambda modes problem of a nuclear power reactor using an h--p finite element method},
    journal = {Annals of Nuclear Energy, pp. 338--349},
    year    = {2014},
--- a/publications-2015.bib
+++ b/publications-2015.bib
@@ -1,4 +1,4 @@
-% Encoding: ISO-8859-1
+% Encoding: US-ASCII
 
 @phdthesis{2015_0,
    author  = {F. S. Alrashed},
--- a/publications-2015.bib
+++ b/publications-2015.bib
@@ -267,7 +267,7 @@
 }
 
 @article{2015_25,
-   author  = {T. Carraro and C. Goll and A. Marciniak-Czochra and A. Mikelic´},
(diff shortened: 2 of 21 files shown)


jppelteret:
The `bib` files for years 2014, 2015 and 2017 appear to contain some incompatible characters... Before this is merged I want to work out where they are so I can replace them.

jppelteret:
Note to self: Jabref is not very helpful about emitting where illegal characters can be found. When you save it just complains. This command is useful to show which line in a file cannot be converted from one format to another:
```sh
iconv --from-code=iso-8859-1 --to-code=us-ascii publications-xxxx.bib
```

publications-2015.bib
```diff
@@ -267,7 +267,7 @@ @article{2015_24
 }
 
 @article{2015_25,
-   author  = {T. Carraro and C. Goll and A. Marciniak-Czochra and A. Mikelic´},
```
davydden: i think this is supposed to be just `\'{c}`, see https://arxiv.org/abs/1401.6278

davydden: but to be fair, i don' know how `\'{c}'` renders...

publications-2014.bib
```diff
@@ -1175,7 +1175,7 @@ @article{2014_109
 }
 
 @article{2014_110,
-   author  = {A. Vidal-Ferrandiza and R. Fayeza and D. Ginestarb and G. Verdúa},
+   author  = {A. Vidal-Ferrandiza and R. Fayeza and D. Ginestarb and G. Verd\'{u}a},
```
davydden: apparently, same `Verdu`, see https://www.sciencedirect.com/science/article/pii/S0306454914002473

davydden: oh, all surnames have extra `a`/`b` here

publications-2015.bib
jppelteret: Oh yeah, thats a mistake... My editor placed an extra character there. Will fix.

publications-2014.bib
jppelteret: Indeed :-)

jppelteret:
@davydden Thanks for pointing out those issues. I think that I've fixed them now.

publications-2014.bib
bangerth: I bet they were copied verbatim from a list of authors where a footnote symbol referred to an address at the bottom of the page. Might have been my fault.

jppelteret: Yip, I agree that this is probably what happened. Although it wasn't a big deal to fix the issue now, I don't think that we should worry too much about this right now. We'll have to go through all of the entries to double check the data anyway.

PR #24, opened 2018-11-23
Add two entries.

bangerth:
These are the only other two that I believe are missing from my ad-hoc file. They were sent to me after I shared that file with @jppelteret and @tjhei offline last.


bangerth:
I saw that in your other entries, you always have "timestamp" and "owner" fields. I'm not sure they add value, but could be convinced otherwise :-)

publications-2018.bib
```diff
@@ -677,4 +677,31 @@ @Article{Chandrashekar2018a
...
+  author = {Perry-Houts, Jonathan and Karlstrom, Leif},
+  title = {Anisotropic viscosity and time-evolving lithospheric instabilities due to aligned igneous intrusions},
+  journal = {Geophysical Journal International},
+  volume = {},
+  number = {},
+  pages = {ggy466},
+  year = {2018},
+  doi = {10.1093/gji/ggy466},
+  URL = {http://dx.doi.org/10.1093/gji/ggy466},
+  eprint = {/oup/backfile/content_public/journal/gji/pap/10.1093_gji_ggy466/2/ggy466.pdf}
```
tjhei: can you remove the eprint line?

tjhei:
> "timestamp" and "owner"

those come from using JabRef and are not useful.

publications-2018.bib
```diff
@@ -677,4 +677,31 @@ @Article{Chandrashekar2018a
...
+@article{doi:10.1093/gji/ggy466,
+  author = {Perry-Houts, Jonathan and Karlstrom, Leif},
+  title = {Anisotropic viscosity and time-evolving lithospheric instabilities due to aligned igneous intrusions},
+  journal = {Geophysical Journal International},
+  volume = {},
+  number = {},
+  pages = {ggy466},
+  year = {2018},
+  doi = {10.1093/gji/ggy466},
+  URL = {http://dx.doi.org/10.1093/gji/ggy466},
```
jppelteret: If the URL points to the same page as the DOI, then this is redundant (since we create a link through the DOI itself). I'd like to leave the URL open to link straight to a PDF (be it preprint, thesis, whatever). I started making this change in the few entries I've edited so far. What do you think? If you agree then we should add this point to the guidelines in the README file.

publications-2018.bib
```diff
@@ -677,4 +677,31 @@ @Article{Chandrashekar2018a
...
+  author =       {Scott Ladenheim, Yi-Chung Chen, Milan D. Mihajlovic, Vasilis},
+  title =        {The {MTA}: {A}n Advanced and Versatile Thermal Simulator for Integrated Systems},
+  journal =      {IEEE Transactions on Computer-Aided Design of Integrated Circuits and Systems},
+  year =         2018,
+  volume =    37,
+  number =    12,
+  pages =     {3123--3136},
+  month =     {January},
+  doi = {10.1109/TCAD.2018.2789729},
+  URL = {http://dx.doi.org/10.1109/TCAD.2018.2789729}}
```
jppelteret: Ditto.

publications-2018.bib
jppelteret: Well, sort of. They generate a clickable link that takes you straight to the page pointed to by `doi.org/<no>`. Take a look at the entry generated for `The deal.II Library, Version 9.0`, that only lists a DOI number. If we added the url `https://www.degruyter.com/view/j/jnma.just-accepted/jnma-2018-0054/jnma-2018-0054.xml` then this would be totally redundant. It'd be nicer to add a URL to the preprint PDF, IMO.

jppelteret: I was talking about the rendered HTML. 

Well, guess I don't really mind if we keep these or not. It looks like I just have a different way of thinking about this than you do.

bangerth:
I ended up removing the URL field. I'm not concerned with anything other than HTML output since nobody in their right mind would want to look at a PDF of this. And since the HTML links to the paper (via its DOI), that's all we care about.

If there happened to be a preprint somewhere, then we could separately link to it via a URL. Though these links are often not permanent, and often also just preprints -- not quite the final paper.

tjhei:
```
ERROR - BibTeX subsystem: /tmp/haoAjph8uy/publications-2018.bib_36.utf8, line 668, syntax error: found "@article", expected end of entry ("}" or ")") (skipping to next "@")
```

publications-2018.bib
```diff
@@ -663,6 +663,30 @@ @PhdThesis{MengyingXiaoPhD2018
...
+
+@article{doi:10.1093/gji/ggy466,
+  author = {Perry-Houts, Jonathan and Karlstrom, Leif},
+  title = {Anisotropic viscosity and time-evolving lithospheric instabilities due to aligned igneous intrusions},
+  journal = {Geophysical Journal International},
+  volume = {},
+  number = {},
+  pages = {ggy466},
+  year = {2018},
+  doi = {10.1093/gji/ggy466}
```
jppelteret: This one should get a `note = {Accepted},`. Having this tag would enable us to more easily locate entries that need to be updated sometime in the future. I'd already added a note to this effect in the README file a few days ago.

publications-2018.bib
```diff
@@ -663,6 +663,30 @@ @PhdThesis{MengyingXiaoPhD2018
...
+}
+
+
+@Article{scott18:mta,
+  author =       {Scott Ladenheim and Yi-Chung Chen and Milan D. Mihajlovi{\'c} and Vasilis F. Pavlidis},
+  title =        {The {MTA}: {A}n Advanced and Versatile Thermal Simulator for Integrated Systems},
+  journal =      {IEEE Transactions on Computer-Aided Design of Integrated Circuits and Systems},
+  year =         2018,
+  volume =    37,
+  number =    12,
```
jppelteret: Each of the `year`, `volume` and `number` entries likely need a to be encapsulated by braces.

PR #26, opened 2018-11-23
Remove HTML framework so that the resulting page can be included into publications.html.

bangerth:
For the moment, let's not merge this yet but discuss whether this is the direction we want to go.

### DIFF
--- a/jabref-template/listrefs.begin.layout
+++ b/jabref-template/listrefs.begin.layout
@@ -1,16 +1,3 @@
-<!DOCTYPE HTML>
-<html>
-  <head>
-    <title>deal.II: Publications</title>
-
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta http-equiv="Content-Type" content="text/html; charset=\encoding">
-    <link rel="icon" href="/favicon.ico"/>
-    <link rel="apple-touch-icon" href="/pictures/clover57.png"/>
-    <meta name="author" content="the deal.II authors"/>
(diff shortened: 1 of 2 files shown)


tjhei:
Yeah, I would prefer we keep things as is because otherwise we can not use/test the html output until we figure out how to merge it and put it online.

jppelteret:
Can we have two templates, with one used for testing and the other for the final render for inclusion into the website?

tjhei:
> Can we have two templates, with one used for testing and the other for the final render for inclusion into the website?

of course!

tjhei:
here is a demo for what I think we can do for the website: we already have a server-side includes, so by moving the static stuff into publications.include and copying the output we have to the webserver, we currently get this: https://dealii.org/temp.html

Before I work on fixing the layout, etc., do we agree that:
- this is all one page (how to cite deal.II and list of publications)? The page is rather long.
- I moved the graph from the sidebar to list further down. This means it is less visible, but it also works on mobile now. I did this mainly to make things easier, though. :-)
- other comments?

bangerth:
The formatting doesn't work right now, but that can wait. I'm ok with leaving the citation info and the publications as is for the moment -- we can split that in a second stage. But I'd like to move the graph to the top again -- I'm ok if you want to put it into the main column, but it should be visible if you get to the page.

tjhei:
> But I'd like to move the graph to the top again

so leave "how to cite deal.II" on the page but move the graph above it? Will do.

jppelteret:
I'm unexpectedly excited about seeing the new publication page go up 🤓

bangerth:
On 3/12/19 2:23 PM, Jean-Paul Pelteret wrote:
> I'm unexpectedly excited about seeing the new publication page go up 🤓

I'm entirely unsurprisingly excited!

bangerth:
No longer relevant following recent patches.

PR #27, opened 2018-11-23
Also print the year in all entries.

bangerth:


tjhei:
It looks like there is a dot after the university, see https://jenkins.tjhei.info/job/publication-list/job/PR-27/2/artifact/output.html
```
N. Dommaraju 
Partition of unity finite element method with overlapping enrichment domains 
. Thesis at Friedrich-Alexander-Universität Erlangen-Nürnberg., 2018.
```
do you have an idea how to remove it?

tjhei:
can you remove the "." in the line that contains ``Thesis``? (two lines up)

jppelteret:
I guess @tjhei was referring to this line:
```
-->\begin{school}. Thesis at \format[HTMLChars]{\school}.\end{school}<!--
```

bangerth:
Fixed that as well.

jppelteret:
Drats. It looks like the leading period might also need to be removed. But what effect would this have on the other entries?

bangerth:
Where do I find the link to the generated HTML?

jppelteret:
I think that the problem here is that the line you'd previously corrected actually has a second period before the word "thesis":
```
-->\begin{school}. Thesis at...
                 ^
```

bangerth:
Oh, I see where the period is coming from. It's because we have this:
```
    \begin{journal} \format[HTMLChars]{\journal}\end{journal}<!--
 -->\begin{booktitle}, In: \format[HTMLChars]{\booktitle}\end{booktitle}<!--
 -->\begin{howpublished}, \howpublished\end{howpublished}<!--
 -->\begin{school}. Thesis at \format[HTMLChars]{\school}\end{school}<!--
```
The problem is that for the entry in question, namely
```
@mastersthesis{2018_14,
   author  = {N. Dommaraju},
   title   = {Partition of unity finite element method with overlapping enrichment domains},
   year    = {2018},
   school  = {Friedrich-Alexander-Universit\"{a}t Erlangen-N\"{u}rnberg},
   url     = {}
}
```
the journal, booktitle, and howpublished entries are empty. So we first print the period of the `school` entry, with nothing before it. 

That's a limitation of that stupid format we have to deal with here. If we had the commas and periods at the end of each string, we'd run into problems if the last entry is empty.

I don't know how to deal with this. We could probably create a `listrefs.mastersthesis.layout` that would catch that and only print fields we expect to exist. I can try that, but would like to do that in a separate PR if that's ok.

jppelteret:
That's fine by me. Should we undo these last two changes and leave them to the next PR, when all is corrected? Or would you prefer to proceed as is?

tjhei:
> I don't know how to deal with this.

You can do conditionals: http://help.jabref.org/en/CustomExports and combine them with logical "and" if needed. It is still a hassle...

bangerth:
Let's just take what we have here.

PR #29, opened 2018-11-24
remove owner and timestamp

### LINKED ISSUE #28: Remove artefacts from JabRef
We could remove the "owner" and "timestamp" fields in the current database. These were generated when the database was first created, and serve no real purpose here.

tjhei:
fixes #28


jppelteret:
Thanks. I must find disable this "feature" in JabRef...

PR #32, opened 2018-11-27
switch pdf to utf8

tjhei:
fixes hundreds of warnings like
```'
WARN - The entry '1998_0' has characters which cannot be encoded in
'ascii'. Recoding problematic characters into macros.
```

### DIFF
--- a/offline/publication_list.tex
+++ b/offline/publication_list.tex
@@ -10,6 +10,8 @@
 %
 
 \documentclass[]{article}
+\usepackage[utf8]{inputenc}
+\usepackage[T1]{fontenc}
 \usepackage[
   backend=biber,
   sorting=ydnt,


tjhei:
> While its true that this fixes these warnings, what it will do is let unicode characters embedded in the bibtex files to be allowed.

No. Check 1998_0. There are no unicode characters in this entry. The point is that biber by default wants to generate utf8 _output_.

jppelteret:
So, what I was going to say is that I think that we can either be conservative and just ignore the warnings (but potentially allow strange characters like ﬂ to creep in) or just hope that nobody makes a PR that does not conform with ASCII encoding.

jppelteret:
> The point is that biber by default wants to generate utf8 output.

But by including these packages with these settings it will pass with a `.bib` file with UTF-8 encoded characters. Previously this failed (which helped me filter out these characters). So this was originally configured in a conservative manner.

jppelteret:
Only Jabref would detect the wrong encoding (because it reads the encoding comment at the top of the bib file), but I don't think that we can't expect everyone contributing something to run their changes through Jabref.

bangerth:
I see that once again I merged something I didn't actually understand :-( I agree with @jppelteret that we should err on the side of requiring canonically correct input. I have no idea how all of the downstream tools we use on these files are going to react to ligatures etc. 

So revert?

jppelteret:
Let me first try to see if I can create a failing case. I'd like to know whether my thinking is correct or not.

bangerth:
I tried, but I can't seem to find any non-ASCII characters in the existing .bib files using the techniques described here: https://stackoverflow.com/questions/3001177/how-do-i-grep-for-all-non-ascii-characters

Is this correct?

jppelteret:
If it's not clear, I used the publications.tex file, in it's original form,
to verify that there were no more unicode characters in the databases.
That's why I advocate that we (maybe) don't want to make this change. I
think I said somewhere in another commit that I know it's a pain for folk
with special characters in their names not to use UTF-8 encoding. But I'm
not sure if the pros of doing so outweigh the potential cons.

bangerth:
We ask folks for bibtex entries anyway -- they should be used to use latex formatting of their names.

jppelteret:
Unless you allow UTF-8 encoding in all of your documents by default, as
many of my European colleagues do. That way they don't have to use old
style latex formatting. So I envision us getting entries sent to us this
way, and the question is whether or not we want to enforce a conservative
style or not. I guess this is up to you and Timo to decide, because you
know what happens when processing it for the webpage, and whatever else you
might wish to use it for. If you do want a clean, portable style why not
let the CI enforce it automatically?

bangerth:
I don't have a strong enough opinion to sway the decision. @tjhei ?

tjhei:
I don't care much either way, I just noticed the biber warnings (see message of this PR) and created a "fix".

My thoughts:
- The main reason for this repo is having the html page and being able to count citations. Everything else like the pdf is not important.
- We can undo this change. JP, do you still think this would be useful?
- We can enforce that no non-ascii characters are present, see #37

jppelteret:
It took me quite some time to sift through all of the character strangeness in the original database that was extracted from the website, and I would just not like for anyone to have decrustify our databases in the future. So I do care, because it was a total waste of time that need not be repeated. I think that #37 is a fix that should work for all parties, so I'm happy with that thanks.

jppelteret:
It goes without saying that I appreciate your listening to my concerns and finding a way to incorporate a fix into the workflow.

PR #35, opened 2018-11-27
Better document and format the style file for readability.

bangerth:


jppelteret:
@bangerth Would you like to rebase this?

jabref-template/listrefs.layout
```diff
@@ -4,27 +4,41 @@
 
 <tr id="\format{\bibtexkey}" class="entry">
   <td>
-    \begin{author}\format[Authors(FirstFirst,Initials,Comma,Comma),HTMLChars]{\author}\end{author}
```
jppelteret: I'm surprised to see the {author} sections go. May I ask, what effect does this have?

jabref-template/listrefs.layout
```diff
@@ -4,27 +4,41 @@
...
+    \format[Authors(FirstFirst,Initials,Comma,Comma),HTMLChars]{\author}
```
bangerth: The author section didn't go. I just don't enclose it in `\begin...\end` any more. That's because *every* entry needs to have an author section, so it's not necessary to make it optional via `\begin...\end`.

The syntax of `\begin{author}  text \end{author}` that jabref offers is weird. It means that `text` is only shown if a bibtex entry has a field called `author`. I'm making the assumption here that *every* entry has a field called `author`.

But nothing bad actually happens if an entry does not have a `author` field: since the `text` here is simply `\format[...]{\author}`, if an entry does not have an `author` field, `text` would just be the empty string. That's not true for some of the text further below where `text` is `, \format{\year}` -- note the comma at the beginning that I definitely only want in the output if there actually is a year.

bangerth:
Rebased.

jabref-template/listrefs.layout
jppelteret: Thanks for the explanation!

PR #36, opened 2018-11-27
Add a template file for PhD theses.

bangerth:
As noted before, theses are printed awkwardly with a leading period. This
can be avoided by having a separate style file for PhD theses. This is a
starting point -- let's see what it renders. If this works, I can add a
similar file for `@mastersthesis` entries.

### DIFF
--- a/jabref-template/listrefs.phdthesis.layout
+++ b/jabref-template/listrefs.phdthesis.layout
@@ -1,0 +1,100 @@
+\begingroup{year}
+  <tr class="show"><td><h3>\format[HTMLChars]{\year}</td></tr></h3>
+\endgroup{year}
+
+<tr id="\format{\bibtexkey}" class="entry">
+  <td>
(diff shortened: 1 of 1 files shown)


jppelteret:
Looking good! I noticed though that the year is not printed anymore:
```
PhD thesis, University of Leicester. 
```
vs
```
. Thesis at University of Leicester, 2018. 
```
Is this intentional?

bangerth:
The missing year should be fixed. (That happened because I had based the branch on something before #27 was merged.)

The second difference you show is due to the fact that I'm only adding a format for PhD theses. The entry with the leading period is marked as `@mastersthesis` and so doesn't run through the template I'm adding here. If this PR looks good and is merged, I'll add another, similar file for `@mastersthesis`.

tjhei:
Formatting looks good. Do we need the "abstract" stuff in there? I see that none of our entries have them, so I guess it doesn't matter...

PR #38, opened 2018-11-28
Fix spelling of 'Texas A&M'.

bangerth:
The name is generally spelled without a space to the left and right of the ampersand.


jppelteret:
Yip, this was almost certainly my fault. Sorry 'bout that!

PR #39, opened 2018-11-28
Add notes for entries that are "not yet published"

jppelteret:


publications-2011.bib
```diff
@@ -121,7 +121,8 @@ @article{2011_10
 @article{2011_11,
    author  = {S. Chandrasekaran and K. R. Jayaraman and M. Gu and H. N. Mhaskar and J. Moffitt},
    title   = {MSNFD: A Higher Order Finite Difference Method for Solving Elliptic PDEs on scattered points (An Extended version of our paper submitted to ICCS 2011)},
-   journal = {submitted},
+   journal = {},
+   note = {Submitted},
```
tjhei: this means the word "submitted" will not show up in the html anymore, right?

publications-2011.bib
```diff
@@ -197,12 +198,13 @@ @phdthesis{2011_18
 @article{2011_19,
    author  = {T. George and A. Gupta and V. Sarin},
    title   = {An empirical analysis of the performance of iterative solvers for SPD systems},
-   journal = {Submitted to ACM Trans. Math. Softw., 2011. IBM Research Report RC 24737},
+   journal = {ACM Trans. Math. Softw., 2011. IBM Research Report RC 24737},
```
tjhei: same here and it is misleading (I assume this never got accepted)

tjhei:
oh gosh, that is a lot of old entries that need to be cleaned up eventually. I assume many got submitted and rejected. :(

jppelteret:
> oh gosh, that is a lot of old entries that need to be cleaned up eventually

Yip. I think that we'll have to go through this with a fine tooth comb.

publications-2011.bib
jppelteret: Hmm... I guess not. It does in the PDF but not the HTML (at least not with our custom output; with Jabref's default export option it did as well).

publications-2011.bib
jppelteret: So how should we proceed here? Would you like me to clean up the entry now or should we do this in another PR?

publications-2011.bib
tjhei: I think we should leave "submitted" inside the journal and just add the "note"

publications-2011.bib
tjhei: I am fine with leaving them as is. I mean, if we have a "note = ..." then it means we need to look at them eventually. And by "we" I don't mean you.

tjhei:
> Yip. I think that we'll have to go through this with a fine tooth comb.

or practically speaking: this is a citation listed by google scholar. We can leave it as is.

publications-2011.bib
bangerth: That's what I had tried to do in #25: to put "submitted" along with the journal. 

But we could add the "note" field to the output. jabref doesn't care whether fields are standardized. If you do 
```
  \begin{note} \format[HTML]{\note} \end{note}
```
it will happily copy the note somewhere.

tjhei: @bangerth can we add this somewhere at the end (maybe even in a separate line)

bangerth: You mean add the note to the HTML output? Yes, it's all possible. I can do that later this afternoon, but let's leave that for a different PR.

tjhei: > You mean add the note to the HTML output?

yes.

> let's leave that for a different PR.

of course!

jppelteret: > You mean add the note to the HTML output?

I like the idea too!

PR #40, opened 2018-11-28
Fix entries for theses.

bangerth:
In preparation for dealing with anything other than PhD theses. I will
want to print the kind of thesis in some way, though I don't know yet.
Either way, it needs to be in a different field than in the school.

I also fix up a few PhD theses. It is not necessary to say 'Doctoral
thesis' -- that's already implied by the '@phdthesis' tag.

Finally, some of the existing thesis entries are misclassified as
phd theses. bibtex only knows two kinds of theses: @phdthesis and
@mastersthesis, so I'm lumping everything that's not a PhD into the
latter category.

### DIFF
--- a/publications-2012.bib
+++ b/publications-2012.bib
@@ -439,7 +439,7 @@
    author  = {K. Kormann},
    title   = {Efficient and Reliable Simulation of Quantum Molecular Dynamics},
    year    = {2012},
-   school  = {Doctoral thesis, Uppsala University},
+   school  = {Uppsala University},
    url     = {http://www.diva-portal.org/smash/record.jsf?pid=diva2%3A549981&dswid=6393}
 }
 
--- a/publications-2013.bib
+++ b/publications-2013.bib
@@ -550,11 +550,12 @@
    doi     = {}
 }
 
-@phdthesis{2013_53,
+@mastersthesis{2013_53,
    author  = {S. H. Lim},
    title   = {Numerical Simulation of Two-phase Heat Flow and Transient Partial Melting in the Lower Crust with Adaptive Mesh Refinement},
    year    = {2013},
-   school  = {Undergraduate honors thesis, University of Michigan, Ann Arbor},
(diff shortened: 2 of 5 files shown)


jppelteret:
> so I'm lumping everything that's not a PhD into the
latter category.

I agree that this is the way to go. Here's how I found __some__ post suggesting that Bachelor theses be documented:
```
@MastersThesis{<bibtex key>,
  document_type = {Bachelor's Thesis},
  author        = {},
  title         = {},
  school        = {},
  year          = {},
  month         = {},
  url           = {},
}
``` 
Now I think that adding the `{Bachelor's/Honours Thesis}` tag to the note field is better than this (maybe invalid, or only valid for `biblatex`?) `document_type` field.

Perhaps you'd be willing to update the README that contains the aforementioned template?

bangerth:
I can put this into README.md, but I'm unclear whether you're advocating for the use of `document_type` or `note` fields? Neither are standard and bibtex will likely ignore them. But jabref still exports them.

jppelteret:
Sorry, I see that my message was a bit ambiguous. `note` is a standard field (see e.g. [this Wikipedia article](https://en.wikipedia.org/wiki/BibTeX#Entry_types); its just marked as "optional" rather than "required"). As best I can tell, these notes do appear in the [PDF output](https://jenkins.tjhei.info/job/publication-list/job/PR-40/2/artifact/offline/publication_list.pdf) (look for "Submitted" and "Accepted" as examples). So I agree that adding this as a `note` might be the best option here; the README just needs to be updated.

bangerth:
How about this?

bangerth:
Oh, I don't know what happened here... I actually made the changes you suggested, but apparently in the wrong branch :-( I didn't mean to make more work for you, sorry...

I'll post them separately, then you can compare with your patch in #43.

jppelteret:
> I actually made the changes you suggested, but apparently in the wrong branch

He he, rookie mistake ;-P

PR #41, opened 2018-11-28
Print 'note' fields.

bangerth:
This should print notes such as 'submitted' or 'accepted' or 'in print'
in parentheses after everything else but before the year. In most cases,
submitted/accepted/in print papers don't have a volume and pages yet,
so this will print the note text in parentheses right after the journal.

There are of course going to be cases where there is no journal. In
that case, '(Submitted), 2018' will be all that's printed. That
is going to look a little funny. If one of you don't like it, any
other suggestions?


tjhei:
> If one of you don't like it, any
> other suggestions?

another option would be to have "Note: <note>" in a separate line below the entry. But I am happy with this as is as well.

PR #44, opened 2018-11-29
Also update the README file about theses.

bangerth:
As mentioned in #40. Competes with #43.

PR #47, opened 2018-11-30
Fix subheading in README

jppelteret:
A subheading was added in #44, but it erroneously appears as a bullet point. This change fixes that.


bangerth:
Oh, yes, I didn't pay attention :-(

PR #57, opened 2019-02-04
Properly capitalize a word.

bangerth:


tjhei:
No, everything is working perfectly, I just disabled the auto-generated status line (and just leave CI, html, latex). This happened while we were building. :-) In short: should work from now on.

PR #60, opened 2019-02-18
Found one of David's publications.

bangerth:


drwells:
I'm impressed that they got my UNC address; we submitted this about a year before I moved!

drwells:
The address should be

Gewerbestrasse 11, 6330 Cham, Switzerland

but otherwise this looks good.

bangerth:
Hm, that seems silly. One generally only lists the city -- like "Springer, London, Paris, Tokyo, Heidelberg".

Or should I just remove the line?

drwells:
You're right. I looked at my bib files and I don't list more than the city when I bother with the address; lets keep it like this.

PR #61, opened 2019-02-26
add plot to html export

tjhei:


tjhei:
it renders it without javascript or external scripts for security reasons. I don't know how to fix it, but it should appear here shortly: https://tjhei.info/jenkins/output.html

PR #66, opened 2019-03-11
Another one.

bangerth:


publications-2018.bib
```diff
@@ -798,4 +798,18 @@ @Article{Choo2018b
   publisher = {Elsevier {BV}},
 }
 
+@article{GDKL18,
+   author = {K. Gupta, Deepak and Keulen, Fred and Langelaar, Matthijs},
+   year = {2018},
+   month = {11},
+   pages = {},
+   title = {Design and analysis adaptivity in multi-resolution topology optimization}
```
drwells: It looks like you are missing a comma here.

bangerth:
Thanks for paying more attention that I apparently did! Fixed version now here!

drwells:
I still think that it is super cool that we can use CI on latex documents like this :)

PR #67, opened 2019-03-12
CI: mark failure status correctly

tjhei:
also fix whitespace while we are here.


tjhei:
I checked that it worked earlier (before removing that commit). I'll just merge, I guess. :-)

PR #68, opened 2019-03-12
add theses from Clemson

tjhei:
I was productive...

### DIFF
--- a/publications-2018.bib
+++ b/publications-2018.bib
@@ -723,6 +723,29 @@
    doi     = {}
 }
 
+@PhdThesis{SergeyCharnyiPhd2018,
+  Title                    = {The {EMAC} scheme for Navier-Stokes simulations, and application to flow past bluff bodies},
+  Author                   = {Sergey Charnyi},
+  School                   = {Clemson University},
+  Year                     = {2018},
+  Url                      = {https://tigerprints.clemson.edu/all_dissertations/2243/}
+}
+
+@PhdThesis{LiangZhaoPhd2018,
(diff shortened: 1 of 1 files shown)


publications-2018.bib
```diff
@@ -723,6 +723,29 @@ @article{GKK18
...
+@PhdThesis{SergeyCharnyiPhd2018,
+  Title                    = {The {EMAC} scheme for Navier-Stokes simulations, and application to flow past bluff bodies},
+  Author                   = {Sergey Charnyi},
+  School                   = {Clemson University},
+  Year                     = {2018},
+  Url                      = {https://tigerprints.clemson.edu/all_dissertations/2243/}
+}
+
+@PhdThesis{LiangZhaoPhd2018,
+  Title                    = {A numerical method for the Navier-Stokes equations in velocity-vorticity form},
```
jppelteret: It doesn't really bother me, but the title of this thesis has different capitalisation on the linked website:
> A Numerical Method for the Navier Stokes Equations in a Velocity-Vorticity Form

drwells: I am not sure if it matters, but the thesis title did not have a hyphen between Navier and Stokes.

tjhei: Sigh.

jppelteret: I think that we need to decide how pedantic we're going to be about these things.

drwells: I cannot imagine a circumstance where the hyphen would matter. Lets not bother with fixing it.

tjhei: No, it makes sense. I was just annoyed that his title should have a hyphen. Thanks guys!

PR #69, opened 2019-03-12
move plot to top of page in html

tjhei:
this is for the dealii.org website. As discussed in #26. See http://dealii.org/temp.html for a demo (in about an hour)

### DIFF
--- a/jabref-template/listrefs.begin.layout
+++ b/jabref-template/listrefs.begin.layout
@@ -421,7 +421,7 @@ p.infolinks { margin: 0.3em 0em 0em 0em; padding: 0px; }
     <!--#include file="publications.include" -->
 
       <div class="row">
-          <div class="well">
+          <div class="well" id="list">
             <h2>Publications using deal.II</h2>
 	    <p>
 	      Below we list all publications we know about. Please get in touch if you think there is anything incorrect or missing!
--- a/jabref-template/listrefs.begin.layout
+++ b/jabref-template/listrefs.begin.layout
@@ -407,6 +407,16 @@ p.infolinks { margin: 0.3em 0em 0em 0em; padding: 0px; }
     <!--#include file="header.include" -->
 
     <div id="content" class="container">
+    <div class="row">
+          <div class="well">
+            <h2>Publications</h2>
+	    <p>
+	      Number of publications using deal.II, see <a href="#list">below</a> for details.
+	    </p>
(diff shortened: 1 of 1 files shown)


drwells:
the temp page looks good to me so I approve :+1:. Lets see what others think.

bangerth:
The layout of the page isn't correct, but I'm not sure whether that's because of the mockup or because the markup in the page is incorrect. We can fix this once we're live.

tjhei:
> The layout of the page isn't correct,

Yeah, sorry. I spent some time looking at the various nested divs, but I couldn't figure it out.

tjhei:
Okay, https://www.dealii.org/publications.html is now online. :-)

jppelteret:
Awesome. Thanks a lot Timo!

bangerth:
@tjhei -- I'm playing with the styles. You left the `publications-old.html` file around. Does it still have any use?

tjhei:
> I'm playing with the styles

only edit publications.include, otherwise the changes need to go into the .begin. and .end.layout files in this repo. Does that make sense?


> You left the `publications-old.html` file around. Does it still have any use?

I left it there for reference so we can compare layout and move everything over, etc.. Can go once we are happy.

bangerth:
On 3/13/19 12:21 PM, Timo Heister wrote:
>     I'm playing with the styles
> 
> only edit publications.include, otherwise the changes need to go into the 
> .begin. and .end.layout files in this repo. Does that make sense?

Ah, crap. So how is publications.html then created, and when? How can I test this?


>     You left the |publications-old.html| file around. Does it still have any use?
> 
> I left it there for reference so we can compare layout and move everything 
> over, etc.. Can go once we are happy.

Just zap this stuff. One can always get it out of the repo, but we don't want 
to end up with unused files littering our repositories...

bangerth:
@tjhei -- ping?

tjhei:
> Ah, crap. So how is publications.html then created, and when? How can I test this?

you can make a PR and jenkins will generate the .html and you can download it. Alternatively, you can run ``make`` when you check out the repo (you need docker).

bangerth:
You mean make a PR on the publication-list repo? The problem with downloading the html is that it uses server-side includes, so it's difficult to test locally. 

My question was more how is the `publications.html` on the website updated? It resides in the magic svn repo, but how does do updated versions get in there?

PR #70, opened 2019-03-18
Adjust layout.

bangerth:
This is an attempt at getting us in the right direction with regard
to the currently pretty broken layout of the page. I don't quite know
yet what the issue is, but I hope to get closer to it in increments.

### DIFF
--- a/jabref-template/listrefs.begin.layout
+++ b/jabref-template/listrefs.begin.layout
@@ -404,59 +404,59 @@ p.infolinks { margin: 0.3em 0em 0em 0em; padding: 0px; }
 
 </head>
 <body>
-    <!--#include file="header.include" -->
+  <!--#include file="header.include" -->
 
-    <div class="bs-header">
-      <div class="container">
-        <br />
-        <h2> Publications on and with <abbr>deal.II</abbr>
-        </h2>
+  <div class="bs-header">
+    <div class="container">
+      <br />
+      <h2> Publications on and with <abbr>deal.II</abbr>
+      </h2>
 
-        <p>
-          This page is an attempt at keeping track of all publications
-          that use and describe deal.II. Keeping it up to date is good
-          for our users (who can see what others have already done),
-          but is also important for keeping our funding agencies
-          appraised how widely used deal.II actually is.
-        </p>
-        
-        <p>
-          If you have published anything,
-          including Diploma, Masters or PhD theses, please
-          <a href="https://github.com/dealii/publication-list/blob/master/README.md">let
+      <p>
+        This page is an attempt at keeping track of all publications
+        that use and describe deal.II. Keeping it up to date is good
+        for our users (who can see what others have already done),
+        but is also important for keeping our funding agencies
+        appraised how widely used deal.II actually is.
+      </p>
+
+      <p>
+        If you have published anything,
+        including Diploma, Masters or PhD theses, please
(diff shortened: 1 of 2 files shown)


jabref-template/listrefs.begin.layout
```diff
@@ -406,27 +406,44 @@ p.infolinks { margin: 0.3em 0em 0em 0em; padding: 0px; }
...
+      <div class="row">
+        <div class="col-xs-12 col-md-9">
+          <div class="well" align="center">
 	    <div id="plot" style="width:600px"></div>
 	    <div id="n_publications"></div>
 	  </div>
-    </div>
-
+        </div>
+      </div>
```
jppelteret: There are three new `<div>`'s in this block, but only two `</div>`'s. Is this perhaps the cause of the build failure?

jabref-template/listrefs.begin.layout
```diff
@@ -406,27 +406,44 @@ p.infolinks { margin: 0.3em 0em 0em 0em; padding: 0px; }
 <body>
     <!--#include file="header.include" -->
 
+    <div class="bs-header">
+      <div class="container">
+        <br />
```
drwells: My browser interprets this as `<br>`. Is `<br />` somehow different? I have not seen this before.

bangerth: They're treated equally. `<br />` is equivalent to `<br></br>` and is the common XML form of a self-closing tag. XML wants every tag to be closed, so a `</br>` is necessary. But nothing useful can be between `<br>` and `</br>`, so `<br />` is the commonly used abbreviation.

tjhei:
Error message is:
```
Could not export file 'source/output.html': java.lang.StringIndexOutOfBoundsException: Backslash parsing error near '</tbody> </table> <footer>  
```

jabref-template/listrefs.end.layout
```diff
@@ -95,7 +95,9 @@
     var n_publications=0;
     for (var i=first_year; i<=last_year; ++i)
       n_publications += table[i-first_year+1][1];
-    document.getElementById("n_publications").innerHTML = "Total number of publications: <b>" + n_publications + "</b>.";
+    document.getElementById("n_publications").innerHTML =
+        "Known publications using deal.II. Total: <b>" + n_publications +
+        "</b>. See <a href=\"#list\">below for individual publications</a>.";
```
tjhei: this ``\"`` might also freak out JabRef.

bangerth:
I indented both the `.begin.layout` and `.end.layout` files for HTML. The `div` elements match correctly now, so it must have been wrong before.

jabref-template/listrefs.end.layout
tjhei: can you try single quotes to see if that fixes the problem? Otherwise move the text outside javascript back into the body.

bangerth: Hm, let me take this out then.

bangerth:
Let's see about this again. I'm pretty sure the `div`s match, and I've removed the escaped double quotes. I can see if we can add the same text again in a follow-up.

If anyone really wants to look through the details of the patch, you may want to do this commit by commit.

tjhei:
This looks good and tests pass! I will merge and copy to the webserver in a minute...

tjhei:
looks great: https://www.dealii.org/publications.html

Thanks!

PR #71, opened 2019-03-18
Remove a 'width:' statement that messes up the page layout.

bangerth:
This statement is responsible for the broken layout of the page. There are
other issues (textual, further layout, etc) that can be addressed in follow-up
PRs, but this is the key to getting the page look halfway reasonable.


bangerth:
Does not supersede #70, which is independent and should still be merged.

bangerth:
Of course, I believe that this isn't automatically propagated to the website right now. So I made the same change by hand to `publications.html` in the svn repo. It will eventually be overwritten.

PR #72, opened 2019-03-19
Further layout improvements.

bangerth:
This PR does two things:
* The first commit moves the table of contents up to the top row of
  boxes. This seems like the better place for it.
* The second commit simply includes the rest of the 'publications.include'
  file into the jabref top-of-the-file template. Right now, the contents
  of the publications.html are cobbled together from a variety of places:
  listrefs.begin.layout, listrefs.end.layout, the publications.include
  file that lives in a different repository, and the bibtex database.
  This just seems too complicated. The different repositories also
  make it difficult to sync changes that would affect files from both
  repos, as I do in the first commit. So I decided to just get rid of
  the publications.include file.

Once merged, and once the new publications.html file shows up in the
svn repository, we should remove the publications.include file there.

### DIFF
--- a/jabref-template/listrefs.begin.layout
+++ b/jabref-template/listrefs.begin.layout
@@ -437,6 +437,20 @@ p.infolinks { margin: 0.3em 0em 0em 0em; padding: 0px; }
 	  <div id="n_publications"></div>
 	</div>
       </div>
+
+      <div class="hidden-xs col-sm-3">
+        <div class="well" style="height:700px">
+          <h2>Contents</h2>
+
+          <p>
+            <a href="#referencing">Referencing <abbr>deal.II</abbr></a>
+            <br><br>
+            <a href="#details">Publications on details
+              of <abbr>deal.II</abbr></a>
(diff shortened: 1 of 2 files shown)


tjhei:
> * So I decided to just get rid of
>   the publications.include file.

sadly jabref doesn't like that:
```
Could not export file 'source/output.html': java.lang.UnsupportedOperationException: bibtex entry fields not allowed in begin or end layout
```

If you still want this, we could move the publications.include into this repo and dump it into the .html during compilation.

bangerth:
Ah, so many limitations :-(

Yes, can we do this? Would you be willing to write up a PR for this? I would then re-do my PR on top of that...

tjhei:
maybe like in #73?

bangerth:
Now rebased on top of #73. The second commit is removed one level of `<div>`s and then indents everything accordingly. There is rather less there than it looks.

tjhei:
okay, looks good. I will merge and publish at dealii.org...

PR #73, opened 2019-03-19
insert publications.include

tjhei:


bangerth:
Well, except it doesn't work:
```
+ sed -n /publications.include/,$$p /home/bob/source/output.html
sed: -e expression #1, char 25: unknown command: `$'
```
You probably need to escape something.

tjhei:
> Ugh, advanced sed. As a PERL person, this hurts. As a matter of fact, I don't understand what these sed commands do, but if they work for you, then that's fine by me.

process ``include publications.include`` in the generated .html file. If you have a more elegant solution for this, please let me know.

bangerth:
On 3/19/19 10:27 AM, Timo Heister wrote:
> process |include publications.include| in the generated .html file. If you 
> have a more elegant solution for this, please let me know.

No. I understood what the commands are doing, just not what the actual syntax is.

PR #77, opened 2019-03-20
Fix layout.

bangerth:
This is a follow-up to #72. The current version looks rather awkward
because I forgot to remove a 'height' statement :-(

@tjhei: Apologies for the extra work.


tjhei:
yes, just saw the current version. :-)

PR #81, opened 2019-04-05
Added reference to deal2lkit.

luca-heltai:


bangerth:
I think we also list these publications in the bibtex files. If it isn't there yet, can you add it?

luca-heltai:
It's already there. :)

PR #85, opened 2019-04-09
one article published

tjhei:


publications-2019.bib
```diff
@@ -143,4 +143,19 @@ @Article{clevenger_par_gmg
...
+  title     = {The numerical study of a microscale model for lithium-ion batteries},
+  journal   = {Computers \& Mathematics with Applications},
+  year      = {2019},
+  month     = {mar},
+  publisher = {Elsevier {BV}},
+  volume    = {77},
+  number    = {6},
+  pages     = {1527--1540},
+  url       = {https://doi.org/10.1016/j.camwa.2018.08.067},
+  doi       = {10.1016/j.camwa.2018.08.067},
```
bangerth: JP mentioned before that it's pointless to have both the URL and the DOI if they point to the same location.

tjhei: true, fixed.

PR #86, opened 2019-04-13
Add two of my publications from 2018

simonsticko:


tjhei:
Thanks! I noticed that your git email address is not associated with your github account. You can add additional email adresses in github or change the commit author email address.

publications-2018.bib
```diff
@@ -886,4 +886,30 @@ @article{kaufl2018
...
+﻿@phdthesis{StickoThesis2018,
+  author = {Sticko, Simon},
+  title = {High Order Cut Finite Element Methods for Wave Equations},
+  school = {Uppsala University},
+  series = {Digital Comprehensive Summaries of Uppsala Dissertations from the Faculty of Science and Technology},
+  ISSN = {1651-6214},
+  number = {1656},
+  ISBN = {978-91-513-0300-0},
+  year = {2018},
+  url={http://urn.kb.se/resolve?urn=urn%3Anbn%3Ase%3Auu%3Adiva-347439}
```
tjhei: I don't think this will work as % is parsed as a comment. Can you try to find a different url?

publications-2018.bib
```diff
@@ -886,4 +886,30 @@ @article{kaufl2018
...
+  month={Aug},
+  volume={76},
+  number={2},
+  pages={812--847},
+  issn={1573-7691},
+  doi={10.1007/s10915-017-0637-y},
+  url={https://doi.org/10.1007/s10915-017-0637-y}
+}
+
+﻿@phdthesis{StickoThesis2018,
```
tjhei: this line contains a non-ASCII character, can you double-check (I think in front of the @)

simonsticko: Oops, strange copy-paste mistake. Fixed it. :)

publications-2018.bib
simonsticko: I'm afraid I can't find one. I can remove the url, but are you sure it's a problem? The entry 2018_47 has a % in the url but the link to that paper works on the website.

tjhei: I take that back, it works correctly. :+1:

PR #102, opened 2019-06-04
Unify the spelling of a number of names.

bangerth:
On the website, it doesn't matter much that we now have some first names
abbreviated in the bibtex entries and others not, since the website
abbreviates all. The point is simply to make writing scripts easier.

### DIFF
--- a/publications-2010.bib
+++ b/publications-2010.bib
@@ -56,7 +56,7 @@
 }
 
 @article{2010_5,
-   author  = {S. Bartle and A. McBride and B. D. Reddy},
+   author  = {S. Bartle and A. T. McBride and B. D. Reddy},
    title   = {Shell finite elements, with applications in biomechanics},
    journal = {In: Kok, S. (Ed.): Proceedings of 7th South African Conference on Computational and Applied Mechanics (University of Pretoria, Pretoria, South Africa)},
    year    = {2010},
@@ -89,7 +89,7 @@
 }
 
 @phdthesis{2010_8,
-   author  = {C. C. Chueh},
+   author  = {C.-C. Chueh},
    title   = {Integrated adaptive numerical methods for transient two-phase flow in heterogeneous porous medi},
    year    = {2010},
    school  = {University of Victoria},
@@ -97,7 +97,7 @@
 }
 
 @article{2010_9,
-   author  = {C. C. Chueh and M. Secanell and W. Bangerth and N. Djilali},
+   author  = {C.-C. Chueh and M. Secanell and W. Bangerth and N. Djilali},
    title   = {Multi-level Adaptive Simulation of Transient Two-phase Flow in Heterogeneous Porous Media},
    journal = {Computers and Fluids},
    year    = {2010},
@@ -149,7 +149,7 @@
 }
 
 @article{2010_14,
-   author  = {M. Freyer and A. Ale and R. Schulz and M. Zientkowska and V. Ntziachristos and K.-H. Englmeier},
+   author  = {M. Freyer and A. Ale and R. B. Schulz and M. Zientkowska and V. Ntziachristos and K.-H. Englmeier},
    title   = {Fast automatic segmentation of anatomical structures in x-ray computed tomography images to improve fluorescence molecular tomography reconstruction},
    journal = {Journal of Biomedical Optics, pp. 036006/1-8},
    year    = {2010},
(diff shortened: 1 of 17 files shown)


publications-2010.bib
```diff
@@ -558,15 +558,15 @@ @article{2010_52
 }
 
 @phdthesis{2010_53,
-   author  = {A. J. Salgado},
+   author  = {A.J. Salgado},
```
tjhei: above you had a space between the first names, is this change on purpose?

publications-2011.bib
```diff
@@ -559,7 +559,7 @@ @article{2011_52
 }
 
 @mastersthesis{2011_53,
-   author  = {D. Riedlbauer},
+   author  = {D Riedlbauer},
```
tjhei: why?

publications-2015.bib
```diff
@@ -1001,7 +1001,7 @@ @phdthesis{2015_95
 }
 
 @article{2015_96,
-   author  = {S. M. Mallikarjunaiah and J. R. Walton},
+   author  = {S. M. Mallikarjunaiah and J. R. J. R. Walton},
```
tjhei: this looks incorrect

publications-2015.bib
```diff
@@ -879,7 +879,7 @@ @article{2015_83
 }
 
 @article{2015_84,
-   author  = {P. A. Kuberry},
+   author  = {Paul Kuberry},
```
tjhei: why do you go to the long form here?

tjhei:
I am not a fan of the mixture of abbreviations and full first names that you introduce here. Of course, we don't really care too much about the details of each entry, so this isn't too bad.

bangerth:
I fixed the issues @tjhei marked up (plus a few more). As for the questions of why I made certain substitutions: I got a list of substitutions that I applied via a script. The issue was that the database contains "W. Bangerth" and "Wolfgang Bangerth" (and similar for other names) in a variety of places, and to make names unique, one has to choose one canonical form. Which one doesn't really matter, and the substitution list I got picked one based on criteria I don't know. So sometimes I expanded short to long, and sometimes long to short form first name.

For the website it doesn't matter because everything there is abbreviated anyway. It's really just for consistency when using other tools on the database.

jppelteret:
I think that you might be able to do this consistently using Jabref. See https://tex.stackexchange.com/a/467688.

bangerth:
On 6/16/19 8:14 AM, Jean-Paul Pelteret wrote:
> I think that you might be able to do this consistently using Jabref. See 
> https://tex.stackexchange.com/a/467688.

Possible -- but the substitutions in this patch were good enough for the 
purpose :-)

jppelteret:
Surely all of this maintenance stuff (i.e. this and subj. of @tjhei's email) is better done using a tool that's specialised for the job? From Jabref's manual:
Cleanup entries: https://help.jabref.org/en/CleanupEntries
Find duplicates: https://help.jabref.org/en/FindDuplicates

jppelteret:
> Possible -- but the substitutions in this patch were good enough for the purpose :-)

But are they? We now have other inconsistencies, e.g. 
- https://github.com/dealii/publication-list/pull/102/files#diff-d5d5ebf54de7c20aaa9478996b5d33b0R130
- https://github.com/dealii/publication-list/pull/102/files#diff-4584a82d63b01033100bc6eb929d462cR1194

jppelteret:
Anyway, if we formally decide how we want all entries to be formatted then we can add this information to the README and I can run Jabref's clean-up tools on all bib files at a later stage. I had always intended to do the latter at some point in time.

jppelteret:
> It's really just for consistency when using other tools on the database

Sorry, I forgot that this was the intended purpose of these changes. I'll stay out of this thread now :-D

bangerth:
I don't disagree with any of what you say -- I just don't want to do the work because it's not important to me, nor for what we do so far with the database.

If others think that this patch is pointless, or a step in the wrong direction (as opposed to a modest step in the right direction), then I'm ok with just closing it.

jppelteret:
I can't help but feel that this PR does go in the wrong direction as it introduces unsustainable changes that are almost certainly inconsistent with the canonical entries listed on doi.org. We've mentioned using doi2bib.org to fetch entries a number of times, but this PR would necessarily mean that the data fetched from there would be changed. This seems a bit counterproductive, and of course this script would have to be run once in a while to make everything consistent again.

Also, how do you actually ensure that the changes it makes are legitimate? From my own bib databases, there are a number of authors of Asian decent that have the same "Englishised" surname plus initials and are definitely not the same person.

> The point is simply to make writing scripts easier.

Reading between the lines, are you trying to get metrics on the number of authors on papers that cite deal.II? If so, then can't we just add a script to this repo that temporarily modifies entries and does the author count?

bangerth:
Let's not worry about it.

The translation table was indeed copied from a script that builds the connection graph I posed on the slack channel a few weeks ago. So the script handles these translations already, I just wanted to make these changes permanent. But let's not worry about it too much or spend too much time on it!

PR #107, opened 2019-06-20
Add two preCICE theses.

### LINKED ISSUE #106: Add another two
From the mailing list:

```
@mastersthesis{2018,
  author    = {Schneider, David},
  title     = {Simulation von Fluid-Struktur-Interaktion mit der Kopplungsbibliothek preCICE},
  school    = {Universität Siegen},
  year      = {2018},
  note      = {Bachelorarbeit}
}
```

```
@mastersthesis{2017,
  author    = {Rave, Kevin},
  title     = {Kopplung von OpenFOAM und deal.II Gleichungslösern mit preCICE zur Simulation multiphysikalischer Probleme},
  school    = {Universität Siegen},
  year      = {2017}
}
```

bangerth:
Fixed #106.

### DIFF
--- a/publications-2017.bib
+++ b/publications-2017.bib
@@ -1280,5 +1280,11 @@
   Url                      = {https://tigerprints.clemson.edu/all_dissertations/1985/}
 }
 
+@mastersthesis{rave2017,
+  author    = {Rave, Kevin},
+  title     = {Kopplung von {OpenFOAM} und {deal.II} {G}leichungsl{\"o}sern mit {preCICE} zur {S}imulation multiphysikalischer {P}robleme},
+  school    = {Universit{\"a}t Siegen},
+  year      = {2017}
+}
 
 @Comment{jabref-meta: databaseType:bibtex;}
--- a/publications-2018.bib
+++ b/publications-2018.bib
@@ -922,4 +922,12 @@
   pages={573-592}
 }
 
+@mastersthesis{schneider2018,
+  author    = {Schneider, David},
+  title     = {Simulation von Fluid-Struktur-Interaktion mit der Kopplungsbibliothek preCICE},
+  school    = {Universit{\"a}t Siegen},
+  year      = {2018},
+  note      = {Bachelorarbeit}
+}
+
 @Comment{jabref-meta: databaseType:bibtex;}


drwells:
We use `\"{o}` for umlauts in other entries. I think we should convert these to that style.

drwells:
IIRC we still have problems with unicode here. If I am wrong then this is good as-is.

bangerth:
I've fixed the two umlauts.

publications-2017.bib
```diff
@@ -1280,5 +1280,11 @@ @PhdThesis{RyanRGrovePhD2017
   Url                      = {https://tigerprints.clemson.edu/all_dissertations/1985/}
 }
 
+@mastersthesis{rave2017,
+  author    = {Rave, Kevin},
+  title     = {Kopplung von OpenFOAM und deal.II Gleichungslösern mit preCICE zur Simulation multiphysikalischer Probleme},
```
drwells:
```suggestion
  title     = {Kopplung von OpenFOAM und deal.II Gleichungsl{\"o}sern mit preCICE zur Simulation multiphysikalischer Probleme},
```

bangerth:
Ah, that's where the `\"o` was -- I had only seen all of the `\"a`s. Thanks for pointing it out -- fixed now!

publications-2018.bib
```diff
@@ -922,4 +922,12 @@ @ARTICLE{Freddi2018573
   pages={573-592}
 }
 
+@mastersthesis{schneider2018,
+  author    = {Schneider, David},
+  title     = {Simulation von Fluid-Struktur-Interaktion mit der Kopplungsbibliothek preCICE},
```
drwells: I am sorry that I keep finding stuff, but we should at least have `{preCICE}` here to get the capitalization correct, right? I think we are more lax on other capitalization issues.

bangerth: I mulled this at the other entry I changed, but then realized that it probably doesn't matter. The issue is that bibtex generally converts everything into lower-case for the titles of articles, and consequently if one wants something capitalized, it needs to be enclosed in `{...}`. 

But bibtex doesn't do that for book titles, which are copied verbatim to latex. I suspect that that's also the case for thesis titles, and that consequently the changes I made in the other entry are pointless (though also harmless).

Let's go with this version and I'll look at how it comes out. If it's wrong, I'll fix it up in a follow-up.

drwells: Sounds good.

bangerth: It looks correct on the website!

PR #110, opened 2019-07-01
Add another paper.

bangerth:
This one has pretty cool pictures :-)

It's only available online right now, so no volume or page numbers. But it has the DOI, so should be easy to find and update later on if we want.

### DIFF
--- a/publications-2019.bib
+++ b/publications-2019.bib
@@ -232,4 +232,14 @@
   journal = {Computer Modeling in Engineering {\&} Sciences}
 }
 
+
+
+@Article{LZ19,
+  author =       {Z. Luo and Y. F. Zhao},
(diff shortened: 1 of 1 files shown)


publications-2019.bib
```diff
@@ -232,4 +232,14 @@ @article{Cheng2019
   journal = {Computer Modeling in Engineering {\&} Sciences}
 }
 
+
+
+@Article{LZ19,
+  author =       {Z. Luo and Y. F. Zhao},
```
drwells: I don't have access to this paper, but in the 'cite as' category I just see 'Y Zhao'. Is it 'Y F Zhao' in the paper?

bangerth:
I can't recall where I had the middle initial from. I've taken it out.

PR #114, opened 2019-08-01
Add 2018 TGS publications

safinenko:


bangerth:
Excellent, thanks for adding these publications!

Issue #118, opened 2019-08-07
additional 2018 entries


tjhei:
before adding these:
1. check that they are valid papers that use deal.II
2. check that they are not already in the .bib files.
3. fix up the publication

```
@Article{Schutz2018,
  author = {Sch{\"{u}}tz, SS},
  title  = {{3D Boundary Element Simulation of Droplet Dynamics in Microchannels: How Droplets Squeeze Through Constrictions and Move in Electric Fields}},
  year   = {2018},
  file   = {:home/heister/.local/share/data/Mendeley Ltd./Mendeley Desktop/Downloaded/Sch{\"{u}}tz - 2018 - 3D Boundary Element Simulation of Droplet Dynamics in Microchannels How Droplets Squeeze Through Constrictions and Move.pdf:pdf},
  url    = {https://infoscience.epfl.ch/record/255307},
}

@Article{Koepf,
  author  = {Koepf, JA and Soldner, D and Gotterbarm, MR and Markl, M},
  title   = {{3D Grainstructure simulation in powder bed additive manufacturing}},
  journal = {researchgate.net},
  file    = {:home/heister/.local/share/data/Mendeley Ltd./Mendeley Desktop/Downloaded/Koepf et al. - Unknown - 3D Grainstructure simulation in powder bed additive manufacturing.pdf:pdf},
  url     = {https://www.researchgate.net/profile/Johannes{\_}Koepf2/publication/325194334{\_}3D{\_}Grainstructure{\_}simulation{\_}in{\_}powder{\_}bed{\_}additive{\_}manufacturing/links/5afd3b52458515e9a59b1dd1/3D-Grainstructure-simulation-in-powder-bed-additive-manufacturing.pdf},
}

@Article{SACCO2018,
  author = {SACCO, FGD},
  title  = {{A 3D adaptive boundary element method for potential flow with nonlinear Kutta condition}},
  year   = {2018},
  file   = {:home/heister/.local/share/data/Mendeley Ltd./Mendeley Desktop/Downloaded/SACCO - 2018 - A 3D adaptive boundary element method for potential flow with nonlinear Kutta condition.pdf:pdf},
  url    = {https://www.politesi.polimi.it/handle/10589/140106},
}

@Article{Arndt,
  author        = {Arndt, Daniel and Kanschat, Guido},
  title         = {{A C1-mapping based on finite elements on quadrilateral and hexahedral meshes}},
  journal       = {arxiv.org},
  year          = {2018},
  month         = {oct},
[...]

PR #121, opened 2019-08-09
Additional 9 entries (Issue #118 part2)

ManaswineeB:


bangerth:
Did you check that all of these publications actually *use* deal.II, as opposed to just citing a deal.II paper? I think that, for example, the last one you added doesn't actually use deal.II.

ManaswineeB:
> Did you check that all of these publications actually _use_ deal.II, as opposed to just citing a deal.II paper? I think that, for example, the last one you added doesn't actually use deal.II.

I didn't realize that some papers are just citing the work. I'll check in and re-do this list and the last list!

bangerth:
Thanks, @ManaswineeB -- let us know once you want us to think about this again!

bangerth:
Let's merge this, and I'll go through it in a follow-up.

bangerth:
Updates are in #128.

PR #130, opened 2019-09-26
More from Timo's list.

bangerth:


publications-2018.bib
```diff
@@ -1243,4 +1242,148 @@ @phdthesis{phdthesis-kauffman
 }
 
 
+@PhdThesis{Khattatov2018,
+  author =       {Eldar Khattatov},
+  title =        {Efficient discretization techniques and domain decomposition
+                  methods for poroelasticity},
+  school =       {University of Pittsburgh},
+  year =         2018},
```
tjhei: missing brace?

bangerth:
Fixed.

Is there a way I can test this locally? Could we add a `Makefile` or something that has a target `make check`?

tjhei:
> Is there a way I can test this locally? Could we add a `Makefile` or something that has a target `make check`?

The Makefile in the root directory generates the publications.html but needs docker installed. Building latex is probably enough. I will add a target, but you can just do
```
               cd offline
               pdflatex -interaction=nonstopmode publication_list.tex
               biber publication_list
               pdflatex -interaction=nonstopmode publication_list.tex
```
https://github.com/dealii/publication-list/blob/master/Jenkinsfile#L69-L72

PR #131, opened 2019-09-26
add make target latex

tjhei:
see #130


bangerth:
Thanks!

Though this doesn't appear to work in finding errors. If I purposefully make a mistake (e.g., remove the opening brace you had pointed out in #130), it still runs through successfully...

tjhei:
> it still runs through successfully...

yeah, I noticed that as well. It works for other errors (missing comma), but biber doesn't fail if you have a missing opening brace. :(

Issue #134, opened 2019-10-02
Update @misc formatting


bangerth:
The `@misc` entries introduced in #133 come out funny on https://dealii.org/publications.html -- just search for "second order ensemble" for example. That's unfortunate, because #133 uses exactly the form for the bibtex entry that is produced by arXiv. We should just update the entry that parses `@misc` entries and that is called by jabref.

PR #137, opened 2019-10-04
Update arXiv output template.

bangerth:
#136 did not do the trick. Let's try this one instead.


bangerth:
Most excellent -- it works now!

(In reference to #134.)

PR #149, opened 2019-10-21
Remove a file that I don't think we need.

bangerth:


tjhei:
I am confused. This is the file that is used to compile the latex version of the CI step.

bangerth:
Oh, oops. I had imagined that it's something that was leftover from an experiment. Disregard please!

PR #154, opened 2019-10-22
fix missing year

tjhei:
The missing year breaks https://www.dealii.org/publications.html
Fix this.


bangerth:
Oh! Is that why the graph at the top current doesn't show up?

tjhei:
> Oh! Is that why the graph at the top current doesn't show up?

yes. :)

PR #156, opened 2019-10-23
Makefile: ignore other bib files

tjhei:
The local make target for the html page would paste all .bib files in
the folder together, not only the ones that are supposed to exist. Fix
this.


bangerth:
I guess we'll have to revisit this in 7981 years. But we can deal with it at that time.

PR #162, opened 2019-10-24
Add an output template for @inproceedings.

bangerth:
Previously, we default template was used, but that looks wonky for some entries.


bangerth:
Ping?

masterleinad:
Did you steal this from somewhere?

bangerth:
No, I'm an honest man. I don't steal!

(The real answer is that we have templates for other types of bibtex entries already in this directory, and I just adapted an existing one to the `@inproceedings` case by looking up which kinds of lines can appear in these entries.)

PR #164, opened 2019-10-27
Add more of the 2017 entries.

bangerth:


publications-2017.bib
```diff
@@ -1,5 +1,102 @@
...
+  Title                    = {{Modelling Fluid Structure Interaction problems using Boundary Element Method}},
+  Author                   = {Giuliani, N},
+  School                   = {Scuola Internazionale Superiore di Studi Avanzati - Trieste},
+  Year                     = {2017},
+  Url                      = {https://iris.sissa.it/handle/20.500.11767/57293}
+}
+
+@InProceedings{Goluchova,
+  Title                    = {Epicyclic oscillations of thick relativistic disks},
+  Author                   = {Goluchov{\'{a}}, K and T{\"{o}}r{\"{o}}k, G},
```
drwells: Is this the same as http://adsabs.harvard.edu/abs/2017bhns.work...47H ? The Harvard link has a lot more authors.

publications-2017.bib
```diff
@@ -1,5 +1,102 @@
...
+}
+
+@article{Grohs2017,
+  doi = {10.5802/smai-jcm.26},
+  url = {https://doi.org/10.5802/smai-jcm.26},
+  year = {2017},
+  publisher = {Cellule {MathDoc}/{CEDRAM}},
+  volume = {3},
+  pages = {219--248},
+  author = {Philipp Grohs and Ralf Hiptmair and Simon Pintarelli},
```
drwells: Should we also do `Grohs, P` etc.?

publications-2017.bib
```diff
@@ -1,5 +1,102 @@
...
+  School = {Karlsruhe Institute of Technology},
+  Url                      = {https://publikationen.bibliothek.kit.edu/1000064791/4189242}
+}
+
+@phdthesis{ubsi_1237,
+	author = {Andreas Kr{\”a}mer},
+	title = {Lattice-{B}oltzmann-{M}ethoden zur {S}imulation inkompressibler {W}irbelstr{\”o}mungen},
+	year = {2017},
+	school = {Universit{\”a}t Siegen},
+	booktitle = {Schriftenreihe des Lehrstuhls für Str{\”o}mungs- und Thermodynamik},
```
drwells: Should there be a book title with a phd thesis?

Rombur:
These lines contain non-ascii characters
```
publications-2017.bib:	booktitle = {Schriftenreihe des Lehrstuhls für Str{\”o}mungs- und Thermodynamik},
publications-2017.bib:	organization ={Institut für Fluid- und Thermodynamik}
```

publications-2017.bib
```diff
@@ -1,5 +1,102 @@
...
+  Title                    = {{High-order hybrid methods using Green's functions and finite elements}},
+  Author                   = {Beams, Natalie N},
+  School                   = {University of Illinois at Urbana-Champaign},
+  Year                     = {2017},
+  Url = {https://www.ideals.illinois.edu/bitstream/handle/2142/98387/BEAMS-DISSERTATION-2017.pdf?sequence=1}
+}
+
+@PhdThesis{brand2017forces,
+  Title                    = {{Forces and Flow of Contractile Networks}},
+  Author                   = {Brand, Christoph Alexander},
```
drwells: Should we do `Brand, C A`?

publications-2017.bib
```diff
@@ -1,5 +1,102 @@
...
+	school = {Universit{\”a}t Siegen},
+	booktitle = {Schriftenreihe des Lehrstuhls für Str{\”o}mungs- und Thermodynamik},	
+	url = {https://dspace.ub.uni-siegen.de/handle/ubsi/1237},
+	organization ={Institut für Fluid- und Thermodynamik}
+}
+
+
+@PhdThesis{beams2017high,
+  Title                    = {{High-order hybrid methods using Green's functions and finite elements}},
+  Author                   = {Beams, Natalie N},
```
drwells: Should we do `Beams, N N`?

publications-2017.bib
masterleinad: Looks like so: https://d-nb.info/1144615712/34
Even if the book is published by the institute.

publications-2017.bib
bangerth: Ah, good eye! I had looked at the linked to work, and missed the additional authors. Lesson: Don't ever trust @tjhei's original list of entries.

publications-2017.bib
bangerth: It's the same to bibtex -- "Lastname, Firstname" and "Firstname Lastname" are the same.

publications-2017.bib
bangerth: No, `@phdthesis` doesn't allow this field: http://bib-it.sourceforge.net/help/fieldsAndEntryTypes.php#phdthesis

bangerth:
I think I've fixed all of the issues. The thing about names doesn't make a difference: we are inconsistent in abbreviating first names, but jabref then consistently abbreviates them for the website. Since we get most entries from doi2bib and arxiv, I'm not interested in making entries consistent in this regard: It's not manageable in the long term -- we just have to take what we get.

publications-2017.bib
drwells: I phrased my question poorly. I really meant to ask if we should abbreviate the first names of the authors like we do elsewhere.

drwells:
Since jabref does the right thing lets leave the first names as-is.

drwells:
It looks like @Rombur's comments have also been addressed (I don't see those lines in this PR anymore) so I will merge.

PR #167, opened 2019-11-05
Add more from Timo's list.

bangerth:
Slowly getting there...

### DIFF
--- a/publications-2016.bib
+++ b/publications-2016.bib
@@ -2224,4 +2224,31 @@ author = "Z. Wang and S. Rudraraju and K. Garikipati"
   url =    {10.1038/npjcompumats.2016.12}
 }
 
+@misc{sticko2016higher,
+    title={Higher Order Cut Finite Elements for the Wave Equation},
+    author={Simon Sticko and Gunilla Kreiss},
+    year={2016},
+    eprint={1608.03107},
+    archivePrefix={arXiv},
+    primaryClass={math.NA}
+}
+
+@misc{gassmoeller2016flexible,
+    title={Flexible and scalable particle-in-cell methods for massively parallel computations},
+    author={R. Gassmoeller and E. Heien and E. G. Puckett and W. Bangerth},
+    year={2016},
+    eprint={1612.03369},
+    archivePrefix={arXiv},
+    primaryClass={math.NA}
+}
+
+@misc{davydov2016hadaptive,
+    title={On the $h$-adaptive PUM and the $hp$-adaptive FEM approaches applied to PDEs in quantum mechanics},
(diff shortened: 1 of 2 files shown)


publications-2016.bib
```diff
@@ -2224,4 +2224,31 @@ @Article{rudraraju16:_mechan
...
 }
 
+@misc{sticko2016higher,
+    title={Higher Order Cut Finite Elements for the Wave Equation},
+    author={Simon Sticko and Gunilla Kreiss},
+    year={2016},
+    eprint={1608.03107},
+    archivePrefix={arXiv},
+    primaryClass={math.NA}
+}
```
drwells: I think that we can update this since it has been published in JSC:
https://link.springer.com/article/10.1007/s10915-019-01004-2

publications-2019.bib
```diff
@@ -500,4 +500,52 @@ @misc{witte2019fast
...
+    author={Mats Kirkes{\ae}ther Brun and Thomas Wick and Inga Berre and Jan Martin Nordbotten and Florin Adrian Radu},
+    year={2019},
+    eprint={1903.08717},
+    archivePrefix={arXiv},
+    primaryClass={math.NA}
+}
+
+@misc{endtmayer2019multigoaloriented,
+    title={Multigoal-oriented optimal control problems with nonlinear PDE constraints},
+    author={Bernhard Endtmayer and Ulrich Langer and Ira Neitzel and Winnifried Wollner and Thomas Wick},
```
drwells: The authors are in a different order here:
https://www.ricam.oeaw.ac.at/files/reports/19/rep19-07.pdf

publications-2016.bib
bangerth: I'm torn what to do in such cases. I have a paper for which the arXiv preprint is far more detailed than the final published version. It's actually the one below, and so I thought it makes sense to list both the preprint and the final version. In the current case, it took the authors 3 years to get it published, so I'm tempted to leave the preprint in as well.

publications-2019.bib
bangerth: Interesting observation, but that's the order reported by arXiv: https://arxiv.org/abs/1903.02799

Interestingly, when I click on the PDF on arXiv, the order is indeed the one from the link in the PDF you link to. I'll go with your order.

publications-2016.bib
bangerth: I'm updating the other one, though.

drwells: Thats a reasonable argument for leaving it with the arxiv link.

publications-2019.bib
drwells: Weird. I didn't notice the inconsistency.

bangerth:
Ping?

PR #173, opened 2019-11-06
Clean up the entries of 2017 and 2018.

bangerth:
In particular, update a few and remove duplicates. Some were in the file
for the wrong year.


tjhei:
oh no: another reduction in 2017 citations...

bangerth:
Yes, and the graph doesn't quite look as we want it to either...

PR #174, opened 2019-11-06
2017 is now finalized and with complete data.

bangerth:
Mark the column in the graph as such.


tjhei:
great! I am a little sad the we were more productive in 2016 it seems...

PR #176, opened 2019-11-12
Update a bunch of 2016 entries.

bangerth:


publications-2016.bib
```diff
@@ -890,26 +810,31 @@ @article{2016_77
...
-   title   = {Phase-field modeling of proppant-filled fractures in a poroelastic medium},
-   journal = {Comp. Meth. Appl. Mech. Engrg.},
-   year    = {2016},
-   volume  = {312},
-   pages   = {509--451},
-   url     = {http://www.sciencedirect.com/science/article/pii/S0045782516300305},
-   doi     = {10.1016/j.cma.2016.02.008}
+@article{Lee2016,
+  doi = {10.1016/j.cma.2016.02.008},
+  url = {https://doi.org/10.1016/j.cma.2016.02.008},
```
tjhei: what was our rule regarding duplication of doi and url?

publications-2016.bib
```diff
@@ -712,15 +640,18 @@ @article{2016_59
...
+  doi = {10.1021/acs.jpcc.6b09775},
+  url = {https://doi.org/10.1021/acs.jpcc.6b09775},
+  year = {2016},
+  month = dec,
+  publisher = {American Chemical Society ({ACS})},
+  volume = {120},
+  number = {49},
+  pages = {27871--27881},
+  author = {Tonghu Jiang and Shiva Rudraraju and Anindya Roy and Anton Van der Ven and Krishna Garikipati and Michael L. Falk},
+  title = {Multiphysics Simulations of Lithiation-Induced Stress in Li1$\mathplus${xTi}2O4 Electrode Particles},
```
tjhei:
```
! Undefined control sequence.
<recently read> \mathplus 
```

publications-2016.bib
bangerth: I'm taking entries verbatim from doi2bib.org, and it has both. I don't think it's reasonable to hand-edit these entries every time -- it's enough work as is :-)

bangerth:
Should be fixed now.

publications-2016.bib
tjhei: I agree that the quality of these entries is not particularly important.

PR #188, opened 2020-02-04
Remove some duplicate publications

kronbichler:
I noticed a few duplicates of my publications (some were preprints and published in 2018/2019, the SISC paper was simply listed twice).


kronbichler:
A shape indeed.

PR #194, opened 2020-03-16
reference 9.1 paper is finally published

tjhei:


drwells:
Does the CI now run automatically here without typing `/rebuild`?

tjhei:
> Does the CI now run automatically here without typing `/rebuild`?

If you add the label before you open the pull request, it build immediately. ;-)

bangerth:
Hallelujah!

PR #196, opened 2020-03-16
replace 9.0 bibtex with link

tjhei:
We don't need the old bibtex here anymore, I think. The link to the
bibtex might be useful, though?


bangerth:
You'll have to rebase.

tjhei:
done

PR #201, opened 2020-05-26
Added dewitt2020prismspf

stvdwtt:


masterleinad:
The CI seems to be stuck here.

bangerth:
I was about to just merge, but it turns out that we already have that publication here: https://github.com/dealii/publication-list/blob/master/publications-2020.bib#L92-L103 So I'll close, but not without saying that we very much appreciate you creating this PR, @stvdwtt !

PR #238, opened 2020-09-18
Added entries without doi/arxiv numbers

juleoc02:


bangerth:
Rather than going back and forth, I've just edited a few entries myself. Let's see what the tester has to say.

/rebuild

bangerth (approved):

PR #244, opened 2020-09-18
cleanup

tjhei:
- delete some duplicate entries
- make bibtex keys unique


publications-2018.bib
```diff
@@ -1822,16 +1811,6 @@ @misc{pearson2018interior
     eprint={1806.05896},
     archivePrefix={arXiv},
     primaryClass={math.OC}
-}
```
masterleinad, line 1825: This should not be deleted.

bangerth (approved):
I'm not going to double-check that the entries you removed really are duplicated. So looks good to me!

PR #246, opened 2020-09-18
Suppress most latex warnings that are irrelevant to the CI checks.

bangerth:


drwells (approved):
:+1:. I cannot think of any other latex parameters that we should adjust.

PR #249, opened 2020-09-21
Adjust some options for the graph of publications.

bangerth:
Specifically: (i) better use the vertical range; (ii) don't show every
year as a label on the x-axis, given that it gets a bit crowded by now.


tjhei (approved):

tjhei:
I hope that we need to adjust the 250 max later this year. :-)

bangerth:
Yes, me too!

PR #250, opened 2020-09-21
Find a better way to limit the size of the graph's y-axis.

bangerth:
Turns out that my last PR doesn't actually work -- the graph object just has its own ideas of how you can set the max size. But there is a different parameter one can set, and I'm now setting this parameter to simply the maximum number of publications we've seen per year. That's also future-proof.

Tested locally to actually work, so I have hope that it will also on the website.


Rombur (approved):

bangerth:
Yes, this is much better :-)

PR #251, opened 2020-09-22
2019 Citations

dklong-csu:
Other article types such as dissertations from 2019 (plus one or two normal articles which I missed previously).


bangerth (approved):

jppelteret (approved):
Yip, those are important!

bangerth:
I fixed a bunch of small problems by direct-edit. Works now, so merging.

PR #259, opened 2020-10-23
Added entry Cobb2020

alex-cobb:
Thanks to the maintainers for this wonderful library.


bangerth (approved):
Awesome -- one more paper in the category "I have no idea about this topic -- who would have guessed that deal.II can be useful for that!"

PR #274, opened 2021-02-22
Add/update some more papers

agrayver:


publications-2021.bib
```diff
@@ -36,5 +36,25 @@ @Article{dealii2019design
...
+  year={2021},
+  publisher={Elsevier}
+}
+
+@article{grayver2021td,
+author = {Grayver, Alexander V. and Kuvshinov, Alexey and Werthm{\"u}ller, Dieter},
+title = {Time-domain modeling of 3-D Earth's and planetary electromagnetic induction effect in ground and satellite observations},
+journal = {Journal of Geophysical Research: Space Physics},
+volume = {129},
+pages = {e2020JA028672},
```
drwells, line 54: Are these pages correct? I don't understand why this is not a pair of numbers.

agrayver: It is in the "published online" stage, where the pages are not known yet, but the publisher gives this thing... (I exported ciation from their website...)

drwells: Weird. Well, if the CI doesn't complain lets go for it.

drwells (approved):

PR #276, opened 2021-03-04
Remove an article that doesn't use deal.II.

bangerth:
I was told that the article doesn't actually show results obtained with deal.II.

/rebuild


peterrum (approved):
Back then we did not support simplex and mixed meshes ;)

PR #279, opened 2021-05-09
Add two of my papers published in 2020

simonsticko:
Should both preprints and published versions be listed here? The published paper in my second commit is already listed as a preprint. Should I remove the preprint or leave it?


bangerth:
If they are essentially the same, then let's remove the preprint.

simonsticko:
Changed the preprint for the published version. I also removed another preprint where the published version is already on the publication list.

PR #284, opened 2021-05-17
Test commit entry for thomas wick, others are in open issue that I ju…

tommeswick:
…st opened. More publications are in an open issue. In case it is easier to insert all references via pull requests directly via github, just let me know. Thanks and best regards, Thomas

### DIFF
--- a/publications-2020.bib
+++ b/publications-2020.bib
@@ -1,5 +1,16 @@
 % Encoding: US-ASCII
 
+@book{Wi20_book,
+      author = "Thomas Wick",
+      title = "Multiphysics Phase-Field Fracture: Modeling, Adaptive Discretizations, and Solvers",
+      year = "2020",
+      publisher = "De Gruyter",
+      address = "Berlin, Boston",
+      isbn = "978-3-11-049739-7",
+      doi = "https://doi.org/10.1515/9783110497397",
+      url = "https://www.degruyter.com/view/title/523232"
+}
+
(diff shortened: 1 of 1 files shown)


masterleinad:
> In case it is easier to insert all references via pull requests directly via github, just let me know.

Sure! If your willing to just create a pull request for the remaining items, that would be much appreciated. :slightly_smiling_face:

masterleinad (approved):

marcfehling:
The bibliography entry looks like this on the website:
image

This seems to be the first `@book` on the publication list. It appears to me that there is one required flag missing, hence the leading comma in the last line.

jppelteret:
@marcfehling, I think that there might be a slight issue with our formatting structure (rather than the bibtex entry itself). IIRC the book and article entries are parsed in the same manner, so to the parser they're missing the `@journal` entry.

Screenshot 2021-05-18 at 05 04 31

bangerth:
Yes, that's in the jabref file that translates from bibtex to html. It's difficult to get this right. Open a github issue about it and I'll eventually get to it.

jppelteret:
> Open a github issue about it and I'll eventually get to it.

Done.

> It's difficult to get this right. 

I think that we had a similar problem with one of the other entry types (master's thesis?) and had to customise the formatting a bit. Hopefully figuring this out won't be as much as a pain as it was the first time round.

PR #286, opened 2021-05-18
twick update pubs 2020

tommeswick:
Worked on my own issue #284 from yesterday and inserted publications myself.

PR #290, opened 2021-05-22
Tommeswick patch 1 corrections

tommeswick:


tommeswick:
Added doi

tommeswick:
Done. Added doi into commit.

jppelteret:
This looks to be a duplicate of https://github.com/dealii/publication-list/pull/286, right? If so, then I think that we can close this PR and just use the other one.

PR #294, opened 2021-07-04
Add a template for books.

### LINKED ISSUE #285: Book entries have a leading comma before the location/publisher
See https://github.com/dealii/publication-list/pull/284#issuecomment-842690099

image
image

bangerth:
May fix #285. Certainly doesn't make it any worse, given that `@book` entries don't have most of the fields that the generic entry uses.

/rebuild

### DIFF
--- a/jabref-template/listrefs.book.layout
+++ b/jabref-template/listrefs.book.layout
@@ -1,0 +1,99 @@
+\begingroup{year}
+  <tr class="show"><td><h3>\format[HTMLChars]{\year}</td></tr></h3>
+\endgroup{year}
+
+<tr id="\format{\bibtexkey}" class="entry">
+  <td>
+    \format[Authors(FirstFirst,Initials,Comma,Comma),HTMLChars]{\author}
+    <br>
+    <b>
+      \format[HTMLChars]{\title}
+    </b>
+
+    <br>
+
+   <!-- The following parts of each entry are all optional and may not
+        be present in all bibtex entries that are processed via the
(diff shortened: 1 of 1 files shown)


jabref-template/listrefs.book.layout
```diff
@@ -0,0 +1,99 @@
...
+        comma. To avoid a space between the previous and current entry
+        (which starts with a comma), we end end each line with a
+        comment-start marker and begin the next line with the
+        comment-end marker. 
+
+        Everything is closed with a period after the last entry.
+
+        -->
+
+    -->\begin{publisher}\format[HTMLChars]{\publisher}\end{publisher}<!--
```
jppelteret, line 31: Thanks for adding this @bangerth! I've had a look at the CI `output.html`, and there's an errant `-->` that appears before the publishers name. 
```
T. Wick
Multiphysics Phase-Field Fracture: Modeling, Adaptive Discretizations, and Solvers
-->De Gruyter, Berlin, Boston, 2020. 
```
Maybe it's the one at the beginning of this line?

bangerth: Ah, good catch. Can you try again?

jppelteret: Fixed! Looks great :-)

jppelteret (approved):
What's there to say... It works perfectly for all of the book entries in our citation list. Thanks for adding this!

bangerth:
Nice, good to have this one issue checked off the list :-)

PR #298, opened 2021-08-16
Add Rigsby's MS thesis.

bangerth:
The defense is going on as about to happen as I write this :-)

/rebuild

### DIFF
--- a/publications-2021.bib
+++ b/publications-2021.bib
@@ -447,6 +447,14 @@ doi = {https://doi.org/10.1029/2020JA028672},
 	journal = {Geological Society, London, Memoirs}
 }
 
+@MastersThesis{rigsby2021,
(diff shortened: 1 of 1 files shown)


drwells:
Is there a URL available?

bangerth:
I don't have one, but that's also because the thesis was only defended today and hasn't actually been submitted yet. I just didn't want to forget about it :-)

drwells:
That's a good idea.

I'd prefer to wait until we have a URL so that we don't have to dig through the university library website to find the thesis later.

bangerth:
OK, let's leave this in limbo then until I have a URL -- sometime later this year I hope.

bangerth:
Thanks for checking, @marcfehling ! I added the URL and rebased to master.

PR #311, opened 2021-09-15
Added citations to various collected publications.

marcfehling:
This is covers most of the publications listed in the second half of our googledoc. I will add the remaining ones in a later PR.

I cleaned up duplicated entries on the way.

Further for each publication, I took the date when it was published online, not the date the actual journal issue got published. The reason is that doi2bib.org prepares entries like this.

### DIFF
--- a/publications-2014.bib
+++ b/publications-2014.bib
@@ -1374,4 +1374,18 @@
    url     = {}
 }
 
+@article{Albert2014,
+  doi = {10.1016/j.bpj.2014.04.036},
+  url = {https://doi.org/10.1016/j.bpj.2014.04.036},
+  year = {2014},
+  month = jun,
+  publisher = {Elsevier {BV}},
+  volume = {106},
+  number = {11},
+  pages = {2340--2352},
+  author = {Philipp~J. Albert and Ulrich~S. Schwarz},
+  title = {Dynamics of Cell Shape and Forces on Micropatterned Substrates Predicted by a Cellular Potts Model},
+  journal = {Biophysical Journal}
+}
+
 @Comment{jabref-meta: databaseType:bibtex;}
--- a/publications-2018.bib
+++ b/publications-2018.bib
@@ -13,20 +13,6 @@
   journal = {Applied Mathematics and Computation}
 }
 
-@article{Liu2018,
-  doi = {10.1007/s00466-018-1662-x},
-  url = {https://doi.org/10.1007/s00466-018-1662-x},
-  year = {2018},
-  month = dec,
-  publisher = {Springer Science and Business Media {LLC}},
-  volume = {64},
-  number = {1},
-  pages = {147--161},
-  author = {Chenchen Liu and Celia Reina},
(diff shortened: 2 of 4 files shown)


marcfehling:
There is an Argentinian publication that causes trouble with invalid characters. I need to figure out how to make them bibtex conform.
```
publications-2019.bib:2295:  title = {Formulação Híbrida de Elementos Finitos de Alta Ordem para a Resolução de Problemas de Reação-Difusão},
publications-2019.bib:2296:  journal = {Mecánica Computacional},
```

bangerth:
Hm, what generally matters is when the journal issue was published. So if you have ACM TOMS, vol. 47, issue 1 (2021), then 2021 should be the year, even if the article had already been published online as "first view" in 2020. I believe that's what doi2bib should also give you. In other words, just take what doi2bib does, without further edits to the year.

bangerth:
As for the invalid characters: latex-style use of special characters works.

marcfehling:
> Hm, what generally matters is when the journal issue was published. So if you have ACM TOMS, vol. 47, issue 1 (2021), then 2021 should be the year, even if the article had already been published online as "first view" in 2020. I believe that's what doi2bib should also give you. In other words, just take what doi2bib does, without further edits to the year.

To give you an example: This article was published online Dec 2018, but the journal issue was published in 2019.

https://link.springer.com/article/10.1007/s00466-018-1662-x
image

https://www.doi2bib.org/bib/10.1007/s00466-018-1662-x
image

In these cases, I went with whatever `doi2bib` returned.

tjhei:
On your last example I would go with 2019, which is the issue date. This is what you see when you click "cite" as well.

tjhei:
To explain in a bit more detail for you and for @bangerth :
doi2bib is often wrong with respect to the publishing date of the volume (which is the one we want). I blame the publishers, because doi2bib queries them according to https://doi2bib.org/about . I assume this is because the articles first appear online, the publishers set the date and then forget to update it once they are actually published.
So, when in doubt, look at the year on shown on the website of the DOI.

marcfehling:
> To explain in a bit more detail for you and for @bangerth :
> doi2bib is often wrong with respect to the publishing date of the volume (which is the one we want). I blame the publishers, because doi2bib queries them according to https://doi2bib.org/about . I assume this is because the articles first appear online, the publishers set the date and then forget to update it once they are actually published.
> So, when in doubt, look at the year on shown on the website of the DOI.

Thanks for the explanation! I adjusted those entries which had this problem and were part of this PR. I would like to avoid going over all entries to double-check them, but let me know if you think it's necessary.

Ci passes after fighting with some non-ascii symbols. So ready for review now!

bangerth:
I vote for just not caring too much. Ultimately, we want to have these records, but in which year we don't care. And because we provide a link, people can get to the publication even if we list a paper in the wrong year.

So my vote would be to just take what the DOI record (as queried by doi2bib) says and not make more work for ourselves than necessary.

bangerth (approved):

bangerth:
Let's see what we get here :-)

PR #316, opened 2021-09-19
Mark 2019 data as 'complete'.

bangerth:
I think we've done our best finding publications in 2019. Let's call it done.

/rebuild


tjhei (approved):
Agreed.

PR #338, opened 2022-02-26
Added RoSchrWi2022 paper

mathmerizing:


marcfehling (approved):
Looks good to me!

Usually we don't add abstracts to bibliography entries. I don't feel strongly about keeping/removing them.

bangerth (approved):

Issue #289, opened 2021-05-20
Update `accepted` publications.


marcfehling:
There are a bunch of publications flagged as `Accepted` that date back to 2008.

We should update them at some point and give them a proper `doi`/`url`.

marcfehling:
While there, we should also update all `arXiv` submissions.

PR #341, opened 2022-02-28
Update `accepted` publications.

marcfehling:
Part of #289.

This covers the `accepted` publications. There are still plenty of publications flagged as `submitted` that need to be updated, too, but let's do this in a separate PR.

---

I couldn't find this entry anywhere. If you google it, you can only find it on our publication page. Shall we remove it?
https://github.com/dealii/publication-list/blob/262a0b897a9008d5cc9b091ded3224f83f41004d/publications-2011.bib#L538-L548

---

This appears to be a talk. I can also only find it on our website and on [LinkedIn](https://za.linkedin.com/in/michael-rapson-95651344). Shall we remove it too?
https://github.com/dealii/publication-list/blob/262a0b897a9008d5cc9b091ded3224f83f41004d/publications-2011.bib#L527-L536

### DIFF
--- a/publications-2008.bib
+++ b/publications-2008.bib
@@ -251,16 +251,14 @@
    url     = {}
 }
 
-@article{2008_23,
-   author  = {M. Kronbichler and G. Kreiss},
-   title   = {A hybrid level--set--Cahn--Hilliard model for two-phase flow},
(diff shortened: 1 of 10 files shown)


marcfehling:
> I agree here, and I agree on removing the publications/talks which seem to not be backed up otherwise. But let us wait for one more opinion here.

I can open up a separate PR for this matter.

tjhei (approved):
I agree as well.

drwells (approved):

PR #342, opened 2022-02-28
Fixed Salvador2022.

marcfehling:
image

### DIFF
--- a/publications-2022.bib
+++ b/publications-2022.bib
@@ -8,7 +8,7 @@
   publisher = {Elsevier {BV}},
   volume = {142},
   pages = {105203},
-  author = {Matteo Salvador and Francesco Regazzoni and Stefano Pagani and Luca Dede{\textquotesingle} and Natalia Trayanova and Alfio Quarteroni},
+  author = {Matteo Salvador and Francesco Regazzoni and Stefano Pagani and Luca Dede' and Natalia Trayanova and Alfio Quarteroni},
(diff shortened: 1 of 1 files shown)


kronbichler (approved):
If I as a speaker of Italian would have had to spell it, I would have used "Dedè" (it's a bit like how non-German speakers would spell the German name Müller, Mueller, Muller, M"uller etc), but for names you never know if the correct grammar is not always the correct spelling of a name. As the author's own home page http://www1.mate.polimi.it/~dede/ , lists the name you choose, that would be clear.

marcfehling:
> As the author's own home page http://www1.mate.polimi.it/~dede/ , lists the name you choose, that would be clear.

The author also used that particular name in the article itself, which is why I chose that particular name.

`doi2bib` added the `\textquotesingle` command, which caused the trouble. This command needs the `textcomp` package, which we probably do not load.

So we could either add the `textcomp` package somewhere, or opt for changing for the `'` ASCII character as proposed in this PR.

marcfehling:
On the other hand, the author is spelled "Dedè" in [this article](https://www.sciencedirect.com/science/article/pii/S0021999122001450?via%3Dihub). We will need to be consistent with his spelling when adding the paper from #344.

bangerth (approved):

bangerth:
Maybe not optimal, but better than nothing. And @drwells discovered the same one recently.

PR #347, opened 2022-05-12
add missing publications

tamiko:


publications-2018.bib
```diff
@@ -2682,4 +2682,18 @@ @mastersthesis{Vundla2018
...
+    author = {Matthias Maier and Andrei Nemilentsau and Tony Low and Mitchell Luskin},
+    doi = {10.1021/acsphotonics.7b01094},
+    journal = {ACS Photonics},
+    number = {2},
+    pages = {544--551},
+    title = {Ultracompact amplitude modulator by coupling hyperbolic polaritons over a graphene-covered gap},
+    url = {https://arxiv.org/abs/1709.06626},
+    volume = {5},
+    year = {2018}
+}
```
bangerth, line 2695: Since it was published in a journal, I would suggest removing the URL to the preprint.

bangerth: same for the other publications

bangerth (changes requested):

publications-2018.bib
tamiko: @bangerth Updated.

bangerth (approved):

PR #361, opened 2022-06-15
Add one.

bangerth:
This is really just a preprint right now, but it has a DOI so I thought I'd add it.

/rebuild

### DIFF
--- a/publications-2022.bib
+++ b/publications-2022.bib
@@ -11,6 +11,16 @@
   publisher={Elsevier}
 }
 
+
+@Article{Zhang2022,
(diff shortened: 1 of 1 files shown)


marcfehling:
Hmm, maybe you can add a `note = {accepted}` in this entry, so that we know that we need to update the entry once it's published in an issue.

marcfehling:
> I don't actually know that it's accepted. I just know that the thing I have here uses deal.II. I don't think I can add anything that I know to be true.

On the top of the article website, it says "This is a preprint article, it offers immediate access but has not been peer reviewed". I agree that a different statement like `note = {preprint}` would be a better choice.



I don't feel strongly about it. To me, it's important that it has a doi. So feel free to merge this PR in its current state.

PR #368, opened 2022-07-31
Add a master's thesis.

bangerth:


marcfehling (approved):
An URL to the thesis would be nice, but I guess hasn't been uploaded somewhere (yet).

bangerth:
Correct. I got the entry from the author of the thesis and he told me that he would provide an updated URL once his university provides him with one. I'd prefer to get this merged as-is and update later.

Issue #344, opened 2022-03-10
Update one paper


bangerth:
This just came out: https://www.sciencedirect.com/science/article/pii/S0021999122001450?via%3Dihub

We have two preprints with the same author list and similar names (parts 1&2). Since they can't clearly both be the same as the one paper linked to above, I vote to keep all three.

marcfehling:
Part of #380.

PR #380, opened 2022-08-30
Update publications from MOX

elauksap:
Should solve #379.

PR #382, opened 2022-09-01
Some housekeeping for 2019 - 2022.

marcfehling:
- Fix `month` entries (see Q13 of [bibtex FAQ](https://ctan.org/pkg/bibtex?lang=en))
- Fix `doi` entries (no urls)
- Remove `url` if information redundant to `doi` field
- Update preprints if published in the meantime. Let `url` still point to arxiv preprint.

There are tons of additional `arxiv` preprints for previous years through which we need to go. Part of #289.

### DIFF
--- a/publications-2020.bib
+++ b/publications-2020.bib
@@ -73,7 +73,6 @@ pages = "113431",
 year = "2020",
 issn = "0045-7825",
 doi = "10.1016/j.cma.2020.113431",
-url = "https://arxiv.org/abs/1902.08112",
 author = "D. Jodlbauer and U. Langer and T. Wick"
 }
 
@@ -123,7 +122,7 @@ number = {},
 pages = {1-20},
 keywords = {gradient-based optimization, monolithic formulation, optimal control, optimal design, unsteady nonlinear fluid-structure interaction},
 doi = {10.1002/nme.6372},
-url = {https://arxiv.org/abs/1910.03424},
+url = {https://onlinelibrary.wiley.com/doi/abs/10.1002/nme.6372},
 eprint = {https://onlinelibrary.wiley.com/doi/pdf/10.1002/nme.6372}
 }
 
@@ -149,7 +148,6 @@ pages = "3001 - 3026",
 year = "2020",
 issn = "0898-1221",
 doi = "10.1016/j.camwa.2020.01.005",
-url = "https://arxiv.org/abs/1903.02799",
 author = "B. Endtmayer and U. Langer and I. Neitzel and T. Wick and W. Wollner"
 }
 
@@ -515,7 +513,6 @@ url="https://www.mdpi.com/2075-163X/10/11/985"
 
 @article{Zhao2020,
   doi = {10.1016/j.cma.2019.112742},
-  url = {https://arxiv.org/abs/1905.00671},
   year = {2020},
   month = apr,
   publisher = {Elsevier {BV}},
@@ -878,7 +875,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{McBride2020,
   doi = {10.1016/j.cma.2020.113320},
-  url = {https://arxiv.org/abs/1909.08695},
   year = {2020},
   month = nov,
   publisher = {Elsevier {BV}},
@@ -965,7 +961,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{Witte2020,
   doi = {10.1515/cmam-2020-0078},
-  url = {https://arxiv.org/abs/1910.11239},
   year = {2020},
   month = nov,
   publisher = {Walter de Gruyter {GmbH}},
@@ -978,7 +973,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{KirkestherBrun2020,
   doi = {10.1016/j.cma.2019.112752},
-  url = {https://arxiv.org/abs/1903.08717},
   year = {2020},
   month = apr,
   publisher = {Elsevier {BV}},
@@ -991,7 +985,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{Endtmayer2020,
   doi = {10.1137/18m1227275},
-  url = {https://arxiv.org/abs/1811.07586},
   year = {2020},
   publisher = {Society for Industrial {\&} Applied Mathematics ({SIAM})},
   volume = {42},
@@ -1004,7 +997,6 @@ doi="10.1002/9781119528609.ch15"
 
 @incollection{Endtmayer2020b,
   doi = {10.1007/978-3-030-55874-1_35},
-  url = {https://arxiv.org/abs/1912.04819},
   year = {2020},
   month = aug,
   publisher = {Springer International Publishing},
@@ -1016,7 +1008,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{Murphy2020,
   doi = {10.1016/j.apnum.2020.08.004},
-  url = {https://arxiv.org/abs/1903.09535},
   year = {2020},
   month = dec,
   publisher = {Elsevier {BV}},
@@ -1029,7 +1020,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{Welper2020,
   doi = {10.1137/19m126356x},
-  url = {https://arxiv.org/abs/1901.01322},
   year = {2020},
   publisher = {Society for Industrial {\&} Applied Mathematics ({SIAM})},
   volume = {42},
@@ -1042,7 +1032,6 @@ doi="10.1002/9781119528609.ch15"
 
 @article{Badia2020,
   doi = {10.1137/20m1328786},
-  url = {https://arxiv.org/abs/1907.03709},
   year = {2020},
   publisher = {Society for Industrial {\&} Applied Mathematics ({SIAM})},
   volume = {42},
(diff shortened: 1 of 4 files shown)


publications-2020.bib
```diff
@@ -124,7 +123,7 @@ @Article{WiWo20
 pages = {1-20},
 keywords = {gradient-based optimization, monolithic formulation, optimal control, optimal design, unsteady nonlinear fluid-structure interaction},
 doi = {10.1002/nme.6372},
-url = {https://onlinelibrary.wiley.com/doi/abs/10.1002/nme.6372},
+url = {https://arxiv.org/abs/1910.03424},
```
bangerth, line 126: Same issue here

publications-2020.bib
```diff
@@ -157,8 +148,8 @@ @Article{EndtNeiLaWiWo20
 pages = "3001 - 3026",
 year = "2020",
 issn = "0898-1221",
-doi = "https://doi.org/10.1016/j.camwa.2020.01.005",
-url = "http://www.sciencedirect.com/science/article/pii/S0898122120300122",
+doi = "10.1016/j.camwa.2020.01.005",
+url = "https://arxiv.org/abs/1903.02799",
```
bangerth, line 152: And here

publications-2020.bib
```diff
@@ -933,10 +876,9 @@ @article{BorregalesRevern2020
   journal = {Computational Geosciences}
 }
 
-
 @article{McBride2020,
   doi = {10.1016/j.cma.2020.113320},
-  url = {https://doi.org/10.1016/j.cma.2020.113320},
+  url = {https://arxiv.org/abs/1909.08695},
```
bangerth, line 881: and here

publications-2021.bib
```diff
@@ -470,13 +446,12 @@ @Article{Araujo2021b
   pages = {110024},
   year = {2021},
   issn = {0021-9991},
-  doi = {https://doi.org/10.1016/j.jcp.2020.110024},
-  url = {https://www.sciencedirect.com/science/article/pii/S0021999120307981}
+  doi = {10.1016/j.jcp.2020.110024},
+  url = {https://arxiv.org/abs/1904.08812}
```
bangerth, line 450: And here

publications-2021.bib
```diff
@@ -671,13 +652,24 @@ @phdthesis{fehn2021phd
   url={https://mediatum.ub.tum.de/1601025}
 }
 
+@article{fehn2021anomalous,
+  doi = {10.1017/jfm.2021.1003},
+  url = {https://arxiv.org/abs/2007.01656},
```
bangerth, line 657: and here

publications-2021.bib
```diff
@@ -921,7 +909,7 @@ @article{heltai2021geometry
   title = {Propagating geometry information to finite element computations},
   volume = {47},
   year = {2021},
-  url = {https://dx.doi.org/10.1145/3468428},
+  url = {https://arxiv.org/abs/1910.09824},
```
bangerth, line 912: and here

publications-2021.bib
```diff
@@ -992,14 +976,62 @@ @article{Comeau2021
...
+  doi = {10.1142/s0218202521500044},
+  url = {https://arxiv.org/abs/1912.03812},
+  year = {2021},
+  publisher = {World Scientific Pub Co Pte Lt},
+  volume = {31},
+  number = {01},
+  pages = {133--175},
+  author = {Andrea Bonito and Ricardo H. Nochetto and Dimitrios Ntogkas},
+  title = {{DG} approach to large bending plate deformations with isometry constraint},
+  journal = {Mathematical Models and Methods in Applied Sciences}
```
bangerth, line 997: here too, the URL can't be right

publications-2021.bib
```diff
@@ -992,14 +976,62 @@ @article{Comeau2021
...
+}
+
+@incollection{Klein2021,
+  doi = {10.1007/978-3-030-57784-1_6},
+  url = {https://arxiv.org/abs/1911.04855},
```
bangerth, line 1002: same here

publications-2021.bib
```diff
@@ -992,14 +976,62 @@ @article{Comeau2021
...
+  publisher = {Springer International Publishing},
+  pages = {165--190},
+  author = {Rebecca Klein and Thomas Schuster and Anne Wald},
+  title = {Sequential Subspace Optimization for Recovering Stored Energy Functions in Hyperelastic Materials from Time-Dependent Data},
+  booktitle = {Time-dependent Problems in Imaging and Parameter Identification}
+}
+
+@article{Liu2021,
+  doi = {10.1016/j.cam.2020.113219},
+  url = {https://arxiv.org/abs/1912.08004},
```
bangerth, line 1013: and here

publications-2021.bib
```diff
@@ -992,14 +976,62 @@ @article{Comeau2021
...
+  volume = {386},
+  pages = {113219},
+  author = {Jie Liu and Matthias M\"{o}ller and Henk M. Schuttelaars},
+  title = {Balancing truncation and round-off errors in {FEM}: One-dimensional analysis},
+  journal = {Journal of Computational and Applied Mathematics}
+}
+
+@article{Anselmann2021,
+  doi = {10.1016/j.matcom.2020.10.027},
+  url = {https://arxiv.org/abs/1912.07426},
```
bangerth, line 1026: and here

publications-2022.bib
```diff
@@ -591,7 +603,18 @@ @MastersThesis{Wik1676116
   number = {22027},
   url    = {http://urn.kb.se/resolve?urn=urn:nbn:se:uu:diva-478600},
   year   = {2022},
-  month  = {June}
+  month  = jun
+}
+
+@incollection{BaMaWaWiWo22,
+  doi = {10.1007/978-3-030-92672-4_8},
+  url = {https://arxiv.org/abs/2006.16566},
```
bangerth, line 611: and here

bangerth (commented):
This wasn't the most exciting patch to sift through :-) But I did, all the way to the bottom, and there are two notes I have:

* I marked up a number of places where you provide a URL for an article where the article has appeared in a journal but the URL points to arXiv. This isn't right. The preprint on arXiv is not guaranteed to be identical to the one posted on the journal's web site, and only the latter is normative. In those places, please remove the URL. The DOI is enough to click through to the article, there is no need for a separate URL -- but in particular, if we provide one, it needs to go to the article, and not a preprint.
* I'm of mixed opinion about the removal of arXiv preprint links if a similar article exists. They are often, but not always identical. For example, I have one paper that did not get accepted as originally written, so we put it onto arXiv. Later, we shrunk the article by half and got it published elsewhere. In cases such as this, it might make sense to keep both on the list. Of course, it is not practically possible (=it is not a useful use of anyone's time) to sort out when an arXiv preprint is or isn't identical to a paper. As a consequence, I don't know what to do -- but I might suggest that it isn't the best use of your time to go through all of the arXiv preprints and find out whether there is a corresponding journal article elsewhere in our list.

marcfehling:
> I marked up a number of places where you provide a URL for an article where the article has appeared in a journal but the URL points to arXiv. This isn't right. The preprint on arXiv is not guaranteed to be identical to the one posted on the journal's web site, and only the latter is normative. In those places, please remove the URL. The DOI is enough to click through to the article, there is no need for a separate URL -- but in particular, if we provide one, it needs to go to the article, and not a preprint.

I removed the URLs to the preprints if the bibtex entry refers to the published manuscript. We had multiple URLs pointing to preprints before this patch, so I thought this was common practice in the publication list. I removed those entries as well to have a coherent list.

> I'm of mixed opinion about the removal of arXiv preprint links if a similar article exists. They are often, but not always identical. For example, I have one paper that did not get accepted as originally written, so we put it onto arXiv. Later, we shrunk the article by half and got it published elsewhere. In cases such as this, it might make sense to keep both on the list. Of course, it is not practically possible (=it is not a useful use of anyone's time) to sort out when an arXiv preprint is or isn't identical to a paper. As a consequence, I don't know what to do -- but I might suggest that it isn't the best use of your time to go through all of the arXiv preprints and find out whether there is a corresponding journal article elsewhere in our list.

I see your point. The case you describe seems to me to be rather the exception. In this patch, I changed arxiv preprints to published manuscripts if I could verify that the authors and the title were similar -- I did not check the contents of these articles in detail.

I like to think of preprints as early-stage research that has been advanced through peer review. Thus personally I consider preprints as duplicate entries, which should be left out if a published manuscript is available, or which should only appear alongside it. Because of this and for cases like the one you described, I liked the idea to connect the published article to the preprints (as we already did for some entries).

What do you think of _adding a separate field_ that contains links to preprints? Since we are using `biber` it shouldn't be too difficult to add such a field. And we can adjust the jabref templates to display hyperlinks as `preprint`.

As a side note: I closed #344 as one of the authors @elauksap confirmed in https://github.com/dealii/publication-list/pull/380#discussion_r960295748 that both preprints indeed became the published article.

bangerth (approved):

bangerth:
On 9/1/22 14:13, Marc Fehling wrote:
> What do you think of /adding a separate field/ that contains links to 
> preprints? Since we are using |biber| it shouldn't be too difficult to add
>  such a field. And we can adjust the jabref templates to display hyperlinks
> as |preprint|.

Just because it is possible doesn't mean it's a good use of anyone's time :-) 
Let the following be a guiding question: Does it make the lives of a 
sufficiently large number of people better if we/you spend time on adding 
these links? I think the answer is probably no. There are better things to do :-)

PR #390, opened 2022-09-02
Removed few more duplicates.

marcfehling:
`biber` complained about a few more duplicates.

~I don't know how it managed to identify them. Have a look at the duplicate entry from the 2017 list. Lots of the fields are similar, but none of them except the `pages` field matches character by character. I'm impressed.~

EDIT: It's actually just checking bibtex keys. There was a matching key in the previous year. It was just coincidence that I found another duplicate.

### DIFF
--- a/publications-2016.bib
+++ b/publications-2016.bib
@@ -1351,28 +1351,6 @@ year = {2016}
    doi     = {}
 }
 
-@article{2016_134,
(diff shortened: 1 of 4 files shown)


marcfehling:
This article made it four times into our publication list :-)

image

bangerth (approved):

PR #391, opened 2022-09-07
Use bibtool to format publication list.

marcfehling:
On the weekend I looked around existing solutions to format `bib` files. I was specifically looking for tools that allow us to sort entries and generate citekeys as this would help maintaining the list. I see the following benefits:
- We won't have trouble with merge conflicts of concurrent PRs as all bibtex entries will be placed at the correct spot
- We can easily identify duplicates if the citekey of each entry is generated uniquely from its fields (e.g. `author`, `title`).
- Globally uniform indentation as eye candy

I found that multiple tools to sort `bib` files exist. `biber` in tool-mode is capable, but translates all entries from the `bibtex` to the `biblatex` format. As we want to maintain compatibility, I instead opted to use the well-established [`bibtool`](http://www.gerd-neugebauer.de/software/TeX/BibTool/en/) ([github](https://github.com/ge-ne/bibtool)), which was developed somewhen in the 90s and is still maintained. It is part of the ubuntu repositories as well as archlinux and possibly others.

So far, I've only added the citekey generation and adjustment of indentation to the bibtool configuration to show you that bibtool really only adjusts the formatting. I want to add the sorting mechanism in a follow-up PR (or add it as a separate commit to this one).

I introduced a Makefile target `sort` so that users can easily apply the formatting. To make sure all PRs conform to the format, I've also added a new `sort` stage to the CI.

What is your opinion?

@tjhei -- If you agree to that mechanism, could you make sure that `bibtool` is installed on the system you run the CI on?


bangerth:
I like the idea, though I will note that that means that everyone who adds an entry also has to have bibtool installed -- this just increases the threshold for people to be able to add entries.

It would be an alternative to just add this to `make sort` and do it periodically, rather than on every pull request.

marcfehling:
> I like the idea, though I will note that that means that everyone who adds an entry also has to have bibtool installed -- this just increases the threshold for people to be able to add entries.

It is true that it increases the threshold, but this puts again more maintenance responsibilities on our side as we will not be able to identify duplicates before merging PRs. Plus, it might still happen that we get merge conflicts of concurring PRs.

This threshold is not negligible, but in my opinon very small as bibtool is practically available in repositories of any OS (even macos).

tjhei:
I think Wolfgang's comment makes sense, but I am not 100% opposed to this PR.

> could you make sure that bibtool is installed on the system you run the CI on?

I will need to update the docker image...

> as bibtool is practically available in repositories of any OS (even macos).

What repository on macOS are you talking about, homebrew?

marcfehling:
> I think Wolfgang's comment makes sense, but I am not 100% opposed to this PR.

I can also remove the CI part, and we can address this in a separate PR.

> What repository on macOS are you talking about, homebrew?

Yes, I found it in [homebrew](https://formulae.brew.sh/formula/bib-tool) but also [macports](https://ports.macports.org/port/BibTool/). Furthermore, it is available on [various linux distributions](https://pkgs.org/search/?q=bibtool) (including ubuntu and thus WSL).

I can prepare a wiki entry for installing `bibtool`.

bangerth (commented):
OK to merge if you remove the changes to the readme.

marcfehling:
> OK to merge if you remove the changes to the readme.

Done. I would appreciate if we would merge this PR after #390 though.

bangerth:
OK, you'll have to rebase now.

marcfehling:
I've added installation instructions for bibtool to the readme. If `make sort` does not find `bibtool`, it will output a link to the installation isntructions in the readme.

This PR is now okay to be merged from my side.

bangerth (approved):

Issue #358, opened 2022-06-15
Response to deal.II publications list update request


KJSchwiebert:
Here is the bibtex entry as requested:

@article{AGGUL2022115052,
title = {NS-ω model for fluid–fluid interaction problems at high Reynolds numbers},
journal = {Computer Methods in Applied Mechanics and Engineering},
volume = {395},
pages = {115052},
year = {2022},
issn = {0045-7825},
doi = {https://doi.org/10.1016/j.cma.2022.115052},
url = {https://www.sciencedirect.com/science/article/pii/S0045782522002663},
author = {Mustafa Aggul and Alexander E. Labovsky and Kyle J. Schwiebert},
}

Alternatively, the work can also be found [here](https://www.sciencedirect.com/science/article/abs/pii/S0045782522002663).

Issue #283, opened 2021-05-17
deal.II publications Thomas Wick


tommeswick:
@book{Wi20_book,
      author = "Thomas Wick",
      title = "Multiphysics Phase-Field Fracture: Modeling, Adaptive Discretizations, and Solvers",
      year = "2020",
      publisher = "De Gruyter",
      address = "Berlin, Boston",
      isbn = "978-3-11-049739-7",
      doi = "https://doi.org/10.1515/9783110497397",
      url = "https://www.degruyter.com/view/title/523232"
}

@unpublished{FaYiWi21_EWCO,
   author = {M. Fan and Y. Jin and T. Wick},
   title = {A quasi-monolithic phase-field description for mixed-mode fracture using predictor-corrector mesh adaptivity},
note = {Engineering with Computers (EWCO), accepted for publication},
year = {2021}
}

@unpublished{Wi21_CMAM,
   author = {T. Wick},
   title = {Dual-weighted residual a posteriori error estimates for a penalized phase-field slit discontinuity problem},
note = {Computational Methods in Applied Mathematics (CMAM), accepted for publication},
year = {2021}
}

@article{FrRiWi21_LocModFE,
title = {LocModFE: Locally modified finite elements for approximating interface problems in deal.II},
journal = {Software Impacts},
volume = {8},
pages = {100070},
year = {2021},
issn = {2665-9638},
doi = {https://doi.org/10.1016/j.simpa.2021.100070},
url = {https://www.sciencedirect.com/science/article/pii/S266596382100018X},
author = {Stefan Frei and Thomas Richter and Thomas Wick},
keywords = {Locally modified finite elements, Fitted finite elements, Interface problems, , Deal.II},
abstract = {We describe the Software package LocModFE, which is an implementation of a locally modified finite element method for an accurate solution of interface problems. The code was originally developed in the finite element library Gascoigne 3d and has now been rewritten in the widerspread library deal.II. This makes the concept of locally modified finite elements accessible to many users all over the world. Applications range from simple Poisson interface problems over multi-phase flows to complex multi-physics problems, such as fluid–structure interactions. Being based on deal.II, it provides plenty of possibilities for future extensions, e.g., parallel computing, multigrid solvers or mesh adaptivity.}
}

@article{EndtLaWi20_smart,
author = {Bernhard Endtmayer and Ulrich Langer and Thomas Wick},
doi = {doi:10.1515/cmam-2020-0036},
[...]

Issue #280, opened 2021-05-10
Missing publications Araújo et. al


ju4nk4:
@article{araujo+wadbro+2021,
author = {Araújo, Juan C. and Wadbro, Eddie},
title = {Shape optimization for the strong directional scattering of dielectric nanorods},
journal = {International Journal for Numerical Methods in Engineering},
volume = {n/a},
number = {n/a},
pages = {},
keywords = {directional scattering, Helmholtz problem, light routing, scattering problem, shape optimization},
doi = {https://doi.org/10.1002/nme.6677},
url = {https://onlinelibrary.wiley.com/doi/abs/10.1002/nme.6677},
eprint = {https://onlinelibrary.wiley.com/doi/pdf/10.1002/nme.6677},
}

@article{araujo+engstrom+2021,
label         = {A2},
title = "On spurious solutions encountered in Helmholtz scattering resonance computations in {$R^d$} with applications to nano-photonics and acoustics",
journal = {Journal of Computational Physics},
volume = {429},
pages = {110024},
year = {2021},
issn = {0021-9991},
doi = {https://doi.org/10.1016/j.jcp.2020.110024},
url = {https://www.sciencedirect.com/science/article/pii/S0021999120307981},
author = {Juan C. {Araújo C.} and Christian Engstr\"om},
keywords = {Plasmon resonance, Acoustic scattering resonances, Nonlinear eigenvalue problems, Helmholtz problem, Leaky modes, Quasi-normal modes},
}

@article{araujo+engstrom+campos+roman+2019,
label         = {A3},
author = {Juan C. {Araújo C.} and Carmen Campos and Christian Engstr\"om and Jose E. Roman},
title = "Computation of scattering resonances in absorptive and dispersive media with applications to metal-dielectric nano-structures",
journal = "Journal of Computational Physics",
volume = "407",
pages = "109220",
year = "2020",
issn = "0021-9991",
doi = "https://doi.org/10.1016/j.jcp.2019.109220",
url = "http://www.sciencedirect.com/science/article/pii/S0021999119309258",
keywords = "Plasmon resonance, Resonance modes, Nonlinear eigenvalue problems, Helmholtz problem, PML, Dispersion analysis, Leaky modes, Resonant states, Quasimodes, Quasi-normal modes",
}

jppelteret:
Thanks for the contributions!

jppelteret:
[...]

Issue #265, opened 2020-12-07
One more


bangerth:
This one: https://www.researchgate.net/publication/346484659_Geometrically_exact_elastoplastic_rods_-determination_of_yield_surface_in_terms_of_stress_resultants

marcfehling:
https://github.com/dealii/publication-list/blob/262a0b897a9008d5cc9b091ded3224f83f41004d/publications-2021.bib#L386-L398

PR #392, opened 2022-09-08
Added 'sort' stage to CI.

marcfehling:
Separated from #391.


tjhei:
Do you still want bibtool installed in the docker image or do we leave this as is for now?

marcfehling:
> Do you still want bibtool installed in the docker image or do we leave this as is for now?

I still think it's handy to have in the docker image and in the CI. We can also wait and see how the new feature is received by the community. So there is no hurry right now :-)

marcfehling:
Couldn't call `make sort` as it would open a shell which is not permitted.

So I tried to copy the instructions. `bibtool` doesn't like the jabref additions in the beginning and end of the files, so the sed commands are necessary. However, this led me to the Jenkinsfile character-escaping hell and I don't know how to get out.
https://gist.github.com/Faheetah/e11bd0315c34ed32e681616e41279ef4

marcfehling:
I found a solution for the sort stage.

Calling `make sort` from a shell didn't work. I did not understand why. I heard that it was a [bug in docker](https://github.com/alpinelinux/docker-alpine/issues/182) that should have been solved in version 20.10. @tjhei -- Which version of docker do you run on the server?
```
make: /bin/sh: Operation not permitted
```

I tried to copy the lines from `make sort` into the Jenkins stage, but I went insane figuring out how to escape special characters within `sed`. So I went with an alternative using a regular `diff` that ignores the first and last lines that contain metainformation, which `bibtool` can not handle.

Please have a look at this solution. I will squash the commits to one once approved.

Alternatively, we could also setup a github-actions script for this.

marcfehling:
This is ready for review by the way. I will remove the last commit that checks the CI once approved.

marcfehling:
ping

marcfehling:
> The problem is that it requires contributors to have installed `bibtool`. What do we gain from requiring people to do that additional step, compared to just doing it outselves every few months?

For the library we do it in the same way with the indent checks, so you can apply the same arguments, too. Continous code maintenance. Detecting duplicates early. Users can check early if their entry is valid (i.e. missing commas). Keeping files readable, as entries are sorted in alphabetical order. Less conflicts in concurrent pull requests: in the past users added entries exclusively in the beginning or end of files, which required frequent rebasing.

`bibtool` is well-established and available in repositories for basically any UNIX operating system as documented in the [readme](https://github.com/dealii/publication-list#installing-bibtool). Users who are not comfortable with the tool can still inform us about their publications with an issue or an email as proposed in the [readme](https://github.com/dealii/publication-list#adding-a-new-publication). Others have done it like this in the past, see for example #358 #283 #280 #265.

The additional stage in the CI will issue a report and informs users on how they need to change their entries. So even when users can't run `bibtool`, they can adopt the changes from the report.

We can teach other users that management tools for bib(la)tex bibliographies like `bibtool`, `bibcure`, and `biber --tool` exist. @gfcas for example got to know `bibtool` after we introduced it with `make sort` in #391. He continued to apply it to his own resources, and even proposed a similar patch like this one in #467.

The price however is, and this applies also to the deal.II library itself, that such an indentation requirement adds additional steps for the user, of which installing the tool is the most tedious, but luckily it remains simple.

This check does not get a 'required' flag, so we can still merge it when the user fails to sort their entries. This way this CI stage can simply inform us that something is wrong in the publication list. We can also introduce it as an entirely separate stage in the CI, so that a misaligned bib file does not affect subsequent stages. @tjhei -- What do you think of this approach? I will update this PR accordingly.

bangerth:
Let's see what others think -- opinions?

marcfehling:
Any opinions on the matter?

Issue #389, opened 2022-09-02
Sort entries in `bib` files.


marcfehling:
`biber` offers the option to sort all entries in `bib` files (see 3.1.7 in [biber doc](https://ctan.org/pkg/biber?lang=en)).

Although this brings no potential benefit for publication lists of previous years, I think this would be quite useful for the current year and upcoming ones. It would be a lot easier to find duplicate entries early if we sort by author for example.

I can also imagine including a check like `indent` from the main repository: We can check if the sorted `bib` file differs from the one in a PR, and request a user to change it. It is also not difficult to add a `sort` target to the makefile.

What do you think?

marcfehling:
Just out of curiosity: If you run `biber --tool publications*.bib` in the main folder, it will create a separate file containing all bibtex entries from all years combined, formatted according to the `biber` default style: [publications-1998_bibertool.bib](https://github.com/dealii/publication-list/files/9481473/publications-1998_bibertool.bib.txt)

We can also provide our own stylefile, similar to clang-format.

marcfehling:
So far, I count 44 duplicates in our publication list. I will have a look a them tomorrow
```
$ grep -r "*1" publications*
publications-2002.bib:@Article{cockburn.kanschat.ea:local*1,
publications-2002.bib:@Article{hartmann.houston:adaptive*1,
publications-2003.bib:@Article{hartmann.houston:goal-oriented*1,
publications-2003.bib:@Article{kanschat:preconditioning*1,
publications-2004.bib:@Article{joshi.bangerth.ea:adaptive*1,
publications-2006.bib:@Article{bangtsson.lund:comparison*1,
publications-2006.bib:@Article{hartmann.houston:symmetric*1,
publications-2006.bib:@Article{joshi.bangerth.ea:non-contact*1,
publications-2009.bib:@PhDThesis{willems:numerical*1,
publications-2009.bib:@Article{ziegler.brendel.ea:investigation*1,
publications-2010.bib:@Article{leicht.hartmann:error*1,
publications-2011.bib:@Article{freddi.royer-carfagni:variational*1,
publications-2011.bib:@Article{hinze.kunkel.ea:model*1,
publications-2012.bib:@Article{carraro.joos.ea:3d*1,
publications-2012.bib:@Article{efendiev.galvis.ea:robust*1,
publications-2012.bib:@Article{hai:numerical*1,
publications-2013.bib:@Article{bangerth.heister.ea:deal-ii*1,
[...]

PR #393, opened 2022-09-09
bibtool: enable sorting

marcfehling:
Part of #389. Follow-up to #391.

This PR enables the `sort` feature of `bibtool`. In the previous PR, we only adjusted the citekeys and the general formatting. Now, we enable the default sorting mechanism, which is [sorting by the citekey](https://github.com/ge-ne/bibtool/blob/97fc3d04655c855ca2794c40303b12fee3fbd273/lib/default.rsc#L113).

I haven't checked if all entries are preserved. But a good indicator that we only move code around is that the number of additions equals the number of deletions (plus the two lines in the configuration file).

### DIFF
--- a/bibtool.rsc
+++ b/bibtool.rsc
@@ -8,3 +8,5 @@ print.indent = 2
 print.line.length = 999
 print.use.tab = off
 print.wide.equal = on
+
+sort = on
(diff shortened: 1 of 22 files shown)


bangerth (approved):

bangerth:
To be clear, the patch doesn't just enable sorting, it actually _does_ the sorting.

PR #398, opened 2022-09-11
Removed same file duplicates.

marcfehling:
I checked all duplicates that occurred in the same file (by searching for `*1` in the citekey).

I provided additional information on potential candidates to distinguish them. I found and removed 17 duplicates this way.

Cleaned up a few things along the way and actually found another phd thesis that uses deal.II.


tjhei (approved):
Sad to see publications go. 😉

marcfehling:
> Sad to see publications go. wink

Only the copycats :)

bangerth (approved):

Issue #397, opened 2022-09-10
broken formatting


tjhei:
these entries are formatted incorrectly starting with ", " in the second line:
![image](https://user-images.githubusercontent.com/1531285/189496652-7fea43c5-2dd4-4e5d-b9bb-e35cb6ef239f.png)

@bangerth You know the formatting schema best. Do you know what is happening here?

bangerth:
Let's look at the first entry. It reads
```
@Misc{jodlbauer.langer.ea:matrix-free,
  author  = {Jodlbauer, Daniel and Langer, Ulrich and Wick, Thomas and Zulehner, Walter},
  title   = {Matrix-free Monolithic Multigrid Methods for Stokes and Generalized Stokes Problems},
  doi     = {10.48550/ARXIV.2205.15770},
  url     = {https://arxiv.org/abs/2205.15770},
  publisher = {arXiv},
  year    = {2022},
  copyright = {Creative Commons Attribution 4.0 International}
}
```
The template for the third line of an entry looks like this:
```
    \begin{howpublished}  \howpublished\end{howpublished}<!--
    -->\begin{publisher}, \format[HTMLChars]{\publisher}\end{publisher}<!--
      -->\begin{address}, \format[HTMLChars]{\address}\end{address}<!--
        -->\begin{pages}, pp. \format[FormatPagesForHTML]{\pages}\end{pages}<!--
         -->\begin{note}  (\format[HTMLChars]{\note})\end{note}<!--
-->\begin{archivePrefix}  \format[HTMLChars]{\archivePrefix}\end{archivePrefix}<!--
 -->\begin{primaryClass}  \format[HTMLChars]{\primaryClass}:\end{primaryClass}<!--
       -->\begin{eprint}\format[HTMLChars]{\eprint}\end{eprint}<!--
         -->\begin{year}, \format[HTMLChars]{\year}\end{year}<!--
                     -->.
```
Each `\begin{xxx}...\end{xxx}` is expanded if the bibtex record contains an `xxx={...}` entry. So for this record, there is no `howpublished=...` line, but there is a `publisher={...}` line and so the first thing that is expanded is the second line of the template, which starts with a comma.

[...]

PR #399, opened 2022-09-12
Fix @misc entries.

bangerth:
See #397. There is no reason to start the second part of the output with a comma -- if the first part exists, make it end in a comma and if it doesn't, then the second just starts the line.

That approach isn't so easy for entries further down the list.

/rebuild


marcfehling (approved):

bangerth:
Ping?

PR #402, opened 2022-09-12
bibtool: style improvements

marcfehling:
Part of #389.

Another resource shipped with `bibtool`, which applies the following style improvements according to the author of the tool:
https://github.com/ge-ne/bibtool/blob/master/lib/improve.rsc
- delete empty fields
- correct page ranges to use the long hyphen `--`
- delete delimiters if the field is purely numerical

All these formatting changes are completely subjective. We do not have to apply all of them and can pick out only a selection of the three.

I don't feel strongly about these changes, but I like that it would make our publication list more consistent.

What are your thoughts? I'm not opposed to a veto :-)


publications-2021.bib
```diff
@@ -548,14 +548,14 @@ @Article{kourakos.harter:simulation
 @InProceedings{kronbichler.fehn.ea:next-generation,
   author  = {Kronbichler, Martin and Fehn, Niklas and Munch, Peter and Bergbauer, Maximilian and Wichmann, Karl-Robert and Geitner, Carolin and Allalen, Momme and Schulz, Martin and Wall, Wolfgang A.},
   title   = {A Next-Generation Discontinuous {G}alerkin Fluid Dynamics Solver with Application to High-Resolution Lung Airflow Simulations},
-  year    = {2021},
-  isbn    = {9781450384421},
+  year    = 2021,
+  isbn    = 9781450384421,
```
marcfehling, line 552: Hmm, this change also adjusts `isbn` numbers. I do not know how `bibtex` reacts to this.

bangerth: Bibtex sees a number and parses it correctly. Bibtex doesn't care what the number *means*.

bangerth (approved):

PR #403, opened 2022-09-12
bibtool: sort order for fields.

marcfehling:
Part of #389.

Second to last resource I would like to incorporate in our publication list. This sorts the fields inside each bibtex entry.
https://github.com/ge-ne/bibtool/blob/master/lib/sort_fld.rsc

Again, this is an entirely subjective change with the goal to have a more uniform publication list. We can adjust the order in which we want to sort the fields ourselves -- this is just the default that the author of bibtool thinks looks nice.

This patch just moves fields around. You can see that the number of line additions is similar to the number of line deletions (with the difference being the 3 lines in the configuration file).

### DIFF
--- a/bibtool.rsc
+++ b/bibtool.rsc
@@ -25,4 +25,7 @@ resource improve
 % translate iso 8859/1 characters to latex
 resource iso2tex
 
+# sort order for fields
+resource sort_fld
+
 sort = on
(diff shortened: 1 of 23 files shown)


marcfehling:
Blocked until #406 is merged.

bangerth (approved):

Issue #405, opened 2022-09-13
Bugs in jabref.


marcfehling:
I noticed two issues when looking for `E. Bängtsson, B. Lund` on https://www.dealii.org/publications.html
![image](https://user-images.githubusercontent.com/18285973/189791138-a2e83cf5-7f10-4cc8-9315-6f4d67a41081.png)

1. Why is the 2006 entry listed as Thesis although it is a `@TechReport`? https://github.com/dealii/publication-list/blob/bdd0b09cb28444304ca31e0e8cc4145f1469588b/publications-2006.bib#L19-L28
2. When clicking on the `BibTeX` button of the 2006 entry, the BibTeX entry for the 2008 publication expands. They both share the same citekey `bangtsson.lund:comparison` now since the introduction of bibtool, which is most likely the case for this behavior. This bug has been around before bibtool, but remained hidden until now. A quick fix is to also add the year into the citekey. I'm working on a fix right now.

bangerth:
About 1: `@TechReport` does not have its own `jabref/listrefs.techreport.layout` file and so jabref routes it through the generic `listrefs.layout` file which looks like this:
```
         \begin{journal}\format[HTMLChars]{\journal}\end{journal}<!--
    -->\begin{booktitle}, In: \format[HTMLChars]{\booktitle}\end{booktitle}<!--
 -->\begin{howpublished}, \howpublished\end{howpublished}<!--
       -->\begin{school}. Thesis at \format[HTMLChars]{\school}\end{school}<!--
      -->\begin{address}, \format[HTMLChars]{\address}\end{address}<!--
       -->\begin{volume}, vol. \volume\end{volume}<!--
       -->\begin{number}(\format[FormatPagesForHTML]{\number})\end{number}<!--
        -->\begin{pages}, pp. \format[FormatPagesForHTML]{\pages}\end{pages}<!--
         -->\begin{note}  (\format[HTMLChars]{\note})\end{note}<!--
    -->\begin{publisher}, \format[HTMLChars]{\publisher}\end{publisher}<!--
         -->\begin{year}, \format[HTMLChars]{\year}\end{year}<!--
                     -->.
```
Apparently it automatically translates the `institution` into a `school` field, which I agree is a bit weird. You also see from the template where the period at the top of the line comes from.

marcfehling:
> Apparently it automatically translates the `institution` into a `school` field, which I agree is a bit weird. You also see from the template where the period at the top of the line comes from.

[...]

PR #406, opened 2022-09-13
bibtool: added year to citekey

marcfehling:
Part of #389. This should fix part 2 of #405.

I've changed the citekey that in starts with the year.

The nice thing about this approach is that the publications are now also sorted by year first, which lets us easily identify articles that are in the wrong file. I moved some articles around in the second commit and found a few duplicates.

### DIFF
--- a/publications-2015.bib
+++ b/publications-2015.bib
@@ -567,6 +567,17 @@
   pages   = {2255--2278}
 }
 
+@InProceedings{2015:hai.bause:adaptive,
+  doi     = {10.1115/imece2015-53265},
+  url     = {https://doi.org/10.1115/imece2015-53265},
+  year    = 2015,
+  month   = nov,
+  publisher = {American Society of Mechanical Engineers},
+  author  = {Bhuiyan Shameem Mahmood Ebna Hai and Markus Bause},
+  title   = {Adaptive Multigrid Methods for Extended Fluid-Structure Interaction ({eXFSI}) Problem: Part I --- Mathematical Modelling},
+  booktitle = {Volume 7B: Fluids Engineering Systems and Technologies}
+}
+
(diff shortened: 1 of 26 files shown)


publications-2015.bib
```diff
@@ -567,6 +567,17 @@ @Article{2015:grieshaber.mcbride.ea:uniformly
   pages   = {2255--2278}
 }
 
+@InProceedings{2015:hai.bause:adaptive,
+  doi     = {10.1115/imece2015-53265},
+  url     = {https://doi.org/10.1115/imece2015-53265},
+  year    = 2015,
+  month   = nov,
```
marcfehling, line 574: The conference happened in Nov 15, but the article was first published online Mar 2016.

publications-2018.bib
```diff
@@ -1,74 +1,5 @@
...
-  eprint  = {1708.01773},
-  archiveprefix = {arXiv},
-  primaryclass = {cs.CE}
-}
-
-@Article{2017:emerson.farrell.ea:computing,
-  doi     = {10.1080/02678292.2017.1365385},
-  url     = {https://doi.org/10.1080/02678292.2017.1365385},
-  year    = 2017,
-  month   = aug,
```
marcfehling: Here and for the following: The journals have been published in 2018, but they have been online already in 2017.

publications-2015.bib
bangerth: As a general rule, please do one thing at a time. It's impossible to look through thousands of lines of changes that are automatically generated if you're also making semantic changes. If you need to make a change like this, do it in a separate patch.

marcfehling: It's in the second commit. I can easily turn it into a separate PR if you like.

bangerth (approved):

PR #407, opened 2022-09-14
Added 'preprint' field.

marcfehling:
Based on the discussion in https://github.com/dealii/publication-list/pull/382#pullrequestreview-1093840819.

I've added a new field `preprint` to the biblatex datamodel and the jabref templates. Here is the idea:
- Users who do not have access to journal publications can now at least have a look at the preprint (insofar as it is public, e.g. on arxiv).
- If we replace preprints with the published articles in the future, we can still connect the preprint to the article in a clear way without loss of information.

The patch manifests as follows:
- In the online publication list, another link pointing to the preprint the will be displayed. No preprint field will be part of the bibtex entry if you click on the `BIBTEX` button to expand it.
```
\begin{preprint}
  [<a href="\format{\preprint}" target="_blank">PREPRINT</a>]
\end{preprint}
```
- In the offline publication list, entries with a preprint field will look like this:
preprint

bibtex and biblatex configurations that do not know about the preprint field will simply ignore them.

Right now, I've added the preprint fields to the release publications.

### DIFF
--- a/jabref-template/listrefs.book.layout
+++ b/jabref-template/listrefs.book.layout
@@ -42,12 +42,16 @@
       [<a href="javascript:toggleInfo('\format{\bibtexkey}','bibtex')">BibTeX</a>]
 
       \begin{doi}
-	[<a href="\format[DOICheck]{\doi}" target="_blank">DOI</a>]
+        [<a href="\format[DOICheck]{\doi}" target="_blank">DOI</a>]
       \end{doi}
 
       \begin{url}
         [<a href="\format{\url}" target="_blank">URL</a>]
       \end{url}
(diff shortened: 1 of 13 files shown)


bangerth (approved):

marcfehling:
The feature works quite nicely also on the online version! :-)
preprint

bangerth:
I didn't pay attention -- can we change `[PREPRINT]` to `[Preprint]`?

marcfehling:
> I didn't pay attention -- can we change `[PREPRINT]` to `[Preprint]`?

Sure. I'll prepare a patch.

PR #422, opened 2022-09-15
Mark 2020 as 'complete'.

marcfehling:


bangerth (approved):

bangerth:
It's disappointing that the 2020 bar doesn't quite reach the 2018 and 2019 bars, but that's just how it is I guess...

marcfehling:
There are still 20 potential duplicates in 2019 and 15 in 2018. With that and the pandemic in mind, I think we should consider 2020  a very successful year for deal.II!

bangerth:
Ah, it's a shame about the duplicates. I'll let you work out how many of these really need to be removed.

PR #425, opened 2022-09-16
Add a master thesis.

simonsticko:


publications-2022.bib
```diff
@@ -661,4 +661,14 @@ @Article{2022:zingaro.fumagalli.ea:geometric
...
 }
 
+@MastersThesis{2022:Holm:cutfem,
+  author  = {Holm, Sigmund Eggen},
+  title   = {Cut finite element methods for partial differential equations on moving domains},
+  school  = {Norwegian University of Science and Technology},
+  year    = 2022,
+  month   = aug,
+  note    = {Masters's Thesis},
+  url     = {https://ntnuopen.ntnu.no/ntnu-xmlui/handle/11250/3013356}
```
marcfehling, line 671:
```suggestion
  url     = {https://hdl.handle.net/11250/3013356}
```
`@MastersThesis` as bibentry is sufficient, there is no need for the note.
Using `handle.net` is a safer way to provide a link (similar to doi).

marcfehling (commented):
Looks good! Thank you for providing the reference.

We just introduced a new feature to sort entries in the `bib` files. Would you mind trying it out by calling `make sort`? You would be the first user to try it out and we would appreciate your feedback!

marcfehling (approved):
Thanks a lot!

PR #426, opened 2022-09-16
Fix 2020thiele.

marcfehling:


peterrum (approved):

marcfehling:
I found a preprint to this article. Let me add it to this entry.

publications-2022.bib
```diff
@@ -574,6 +566,15 @@ @Article{2022:stein.comeau.ea:numerical
   url     = {https://www.sciencedirect.com/science/article/pii/S0040195122000701}
 }
 
+@InProceedings{2022:thiele.wick:space-time,
+  author  = {J. P. Thiele and T. Wick},
+  title   = {Space-time error control using a partition-of-unity dual-weighted residual method applied to low mach number combustion},
+  booktitle = {Proceedings of ICOSAHOM, 2020+1},
```
bangerth, line 572: Is this really the year, 2020+1?

If so, please merge yourself.

bangerth (approved):

publications-2022.bib
marcfehling: I've checked their website: https://www.icosahom2020.org/proceedings

The conference was supposed to happen in 2020, but because of the pandemic they moved it to 2021.
image

So the conference actually happened in 2021. Let me move the entry to the 2021 year then.

bangerth (approved):

PR #433, opened 2022-09-19
Added publications of Freese.

marcfehling:


publications-2020.bib
```diff
@@ -657,6 +657,17 @@ @Article{2020:fischer.min.ea:scalability
   publisher = {{SAGE} Publications}
 }
 
+@Article{2020:freese:numerical,
+  author  = {Freese, Jan Philip},
+  title   = {Numerical homogenization of time-dependent Maxwell's equations with dispersion effects},
+  year    = 2020,
```
marcfehling, line 663: The exam was in 2020, but the manuscript was published 2021. To which year do we assign these theses?

peterrum (approved):
I guess 2021 makes more sense.

publications-2022.bib
```diff
@@ -168,6 +168,21 @@ @Misc{2022:fehling.bangerth:algorithms
   copyright = {arXiv.org perpetual, non-exclusive license}
 }
 
+@Article{2022:freese:heterogeneous,
+  author  = {Philip Freese},
```
peterrum, line 172:
```suggestion
  author  = {Freese, Philip},
```
?

marcfehling: bibtex understands it both ways. I agree that this way it is clearer.

publications-2021.bib
```diff
@@ -353,6 +353,17 @@ @Article{2021:fraters.billen:on
...
+@Article{2021:freese:numerical,
+  author  = {Freese, Jan Philip},
+  title   = {Numerical homogenization of time-dependent Maxwell's equations with dispersion effects},
+  year    = 2021,
+  doi     = {10.5445/IR/1000129214},
+  url     = {https://publikationen.bibliothek.kit.edu/1000129214},
+  language = {en},
+  publisher = {Karlsruher Institut f\"{u}r Technologie (KIT)},
+  copyright = {Open Access}
+}
```
bangerth, line 365: This should be marked up as `@PhDThesis`, not `@Article`. It needs to use `school=` instead of `publisher=`.

marcfehling: True. I didn't pay attention. I will fix it.

bangerth (approved):

PR #438, opened 2022-09-22
Add first 4 from the 2021 list.

bangerth:
Two of those are actually in 2020.

### DIFF
--- a/publications-2020.bib
+++ b/publications-2020.bib
@@ -324,17 +324,6 @@
   preprint = {https://arxiv.org/abs/1903.08717}
 }
 
-@Article{2020:bruna-rosso.mergheim.ea:finite,
-  author  = {Claire Bruna-Rosso and Julia Mergheim and Barbara Previtali},
-  title   = {Finite element modeling of residual stress and geometrical error formations in selective laser melting of metals},
-  journal = {Proceedings of the Institution of Mechanical Engineers, Part C: Journal of Mechanical Engineering Science},
-  year    = 2020,
-  pages   = 095440622094322,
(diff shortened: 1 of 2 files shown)


peterrum (approved):

tjhei:
The list that I generate is by filtering google scholar "cited by" results by year. I am sure google scholar gets it wrong many times, so I would not rely on that information.

I would go with 2021 for the first article as this is how the journal picks the year:
```
1.
Bruna-Rosso C, Mergheim J, Previtali B. Finite element modeling of residual stress and geometrical error formations in selective laser melting of metals. Proceedings of the Institution of Mechanical Engineers, Part C: Journal of Mechanical Engineering Science. 2021;235(11):2022-2038. doi:10.1177/0954406220943225
```

marcfehling:
> I would go with 2021 for the first article as this is how the journal picks the year:

Same for the other 2020 article [according to the journal](https://epubs.siam.org/action/downloadCitation?format=bibtex&include=abs&direct=true&doi=10.1137/20M1341210&downloadFileName=siam_20M1341210) which shows 2021.

bangerth:
So fixed.

I will note that I do not spend the time to look up every single entry anyone submits here to determine whether it is correct or not. This is not worth my time, nor do I think that it is worth any of y'all's time. We're trying to provide a reasonable service to ourselves and others in creating these entries, not a perfect one. If doi2bib tells me that the date is 2020, I will simply go with it rather than double and triple check these entries. It is not worth worrying about these things.

marcfehling (approved):

PR #443, opened 2022-09-23
Added 2021 publications K-M.

marcfehling:


marcfehling:
The TeX distribution on the CI does not like the Ā character, which we set with `\=A`.
```
! Package inputenc Error: Unicode char \u8:Ā not set up for use with LaTeX.
```

I don't encounter this problem on my machine. Is it time to update the CI? Alternatively, I can turn the character into A without the bar.

marcfehling:
Let me move the troubled entry into a separate PR.

bangerth (approved):

PR #445, opened 2022-09-24
Update three publications of Wick.

marcfehling:
The looks like the proceedings will be published in 2023:
http://www.ddm.org/DD26/Proc-26.php

### DIFF
--- a/publications-2021.bib
+++ b/publications-2021.bib
@@ -513,26 +513,6 @@
   publisher = {Elsevier {BV}}
 }
 
-@InProceedings{2021:d-jodlbauer:efficient,
-  author  = {D. Jodlbauer, U. Langer, T. Wick},
-  title   = {Efficient monolithic solvers for fluid-structure interaction applied to flapping membranes},
-  booktitle = {Domain Decomposition Methods in Science and Engineering XXVI},
-  year    = 2021,
-  pages   = {accepted},
-  address = {Cham},
-  publisher = {Springer International Publishing}
-}
-
-@InProceedings{2021:d-jodlbauer:parallel,
-  author  = {D. Jodlbauer, U. Langer, T. Wick},
-  title   = {Parallel domain decomposition solvers for the time harmonic Maxwell equations},
-  booktitle = {Domain Decomposition Methods in Science and Engineering XXVI},
-  year    = 2021,
-  pages   = {accepted},
-  address = {Cham},
-  publisher = {Springer International Publishing}
-}
-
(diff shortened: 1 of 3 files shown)


publications-2020.bib
```diff
@@ -217,6 +217,23 @@ @Article{2020:benbow.kawama.ea:coupled
...
+  editor  = {Susanne C. Brenner and Eric Tsz Shun Chung and Axel Klawonn and Felix Kwok and Jinchao Xu and Jun Zou},
+  booktitle = {Domain Decomposition Methods in Science and Engineering XXVI},
+  year    = 2020,
+  series  = {Lecture Notes in Computational Science and Engineering},
+  volume  = 145,
+  pages   = {615--622},
+  address = {Hong Kong, China},
+  month   = dec,
+  publisher = {Springer International Publishing},
+  note    = {In press},
```
bangerth, line 232: Since it is in a published volume, with page numbers, it cannot be "in press" -- it has already been printed. "In press" means it is accepted and has gone through the editing process, but has not been published in print. It is today equivalent to "published online" with no volume/issue/pages associated yet. Just remove the note.

Same below.

bangerth (changes requested):

publications-2020.bib
marcfehling: The volume is not published yet. In fact, the Springer website shows that its copyright year is 2023. I don't see a way to order a hardcopy of the proceedings yet:
https://link.springer.com/book/9783030950248
http://www.ddm.org/DD26/Proc-26.php

Hence I thought `in press` was fitting here. We will need to go over these entries again next year once the volume is actually available.

marcfehling:
Same applies for `2021:thiele.wick:space-time*1` and the not yet published proceedings book:
- https://github.com/marcfehling/publication-list/blob/efaea0f43fcdf2e3a88223185e8c9ad5757822fb/publications-2021.bib#L2126-L2133
- https://link.springer.com/book/9783031204319

bangerth:
What did we resolve to do?

marcfehling:
My approach was to wait until Springer finally releases the proceedings. It looks like there has been progress on their website, i.e. you can finally pre-order one of them.

I would update this PR once the individual articles can be referred to the official Springer page, so the volume, page number, doi and url are correct. We can then add the articles from the conference websites as preprint-links.

marcfehling:
Springer finally published the proceedings this month. I've updated the PR.

bangerth (approved):

PR #447, opened 2022-09-24
Unicode :-(

marcfehling:
See also https://github.com/dealii/publication-list/pull/443#issuecomment-1256704541

The TeX distribution on the CI does not like the Ā character, which we set with `\=A`.
```
! Package inputenc Error: Unicode char \u8:Ā not set up for use with LaTeX.
```
I don't encounter this problem on my machine when running make latex.
image

Is it time to update the CI? Alternatively, I can turn the character into A without the bar.

### DIFF
--- a/publications-2021.bib
+++ b/publications-2021.bib
@@ -1200,7 +1200,7 @@
 
 @Article{2021:karban.panek.ea:fem,
   author  = {Pavel Karban and David P{\'a}nek and Tam{\'a}s Orosz and Iveta Petr{\'a}{\v{s}}ov{\'a} and Ivo Dole{\v{z}}el},
-  title   = {{FEM} based robust design optimization with Agros and Artap},
+  title   = {{FEM} based robust design optimization with Agros and {\=A}rtap},
   journal = {Computers {\&} Mathematics with Applications},
   year    = 2021,
   volume  = 81,


drwells:
I see

```
This is pdfTeX, Version 3.1415926-2.5-1.40.14 (TeX Live 2013/Debian)

 restricted \write18 enabled.

entering extended mode

(./publication_list.tex

LaTeX2e <2011/06/27>
```
in the logs - we are running a 9 year old version of TeX Live. I'll check if my version of `inputec` can actually solve these problems. If so we should upgrade.

drwells:
Yup - it looks like this works for both of us with newer versions of tex live.

bangerth (approved):

bangerth:
Let's not agonize over it and just spell it like a regular "A". The list isn't an attempt to get everything unambiguously right, but instead to help people find out about (and find) papers on specific topics, and they will be able to do that with or without the bar over the A.

bangerth:
Can I again advocate for a "good enough" solution as mentioned above?

bangerth:
Ping?

marcfehling:
I've added the actual article in #492. This PR now only changes the character in question.

We might run into other problems caused by old Latex versions in the future, so I would vote for leaving it open until the Docker image got an update. Alternatively, we can let github-actions perform the Latex check, or do we use the generated pdf for something else?

tjhei:
I can update the image at some point. I am in the process of moving and updating the Jenkins server, but updating the main system is of course of higher priority for me.
The artifacts (pdf and html) are linked in the readme and automatically downloaded on the webserver, so switching to GitHub actions would require some changes.

bangerth:
Folks, we're talking about replacing an 'a' for an 'a' and you're proposing a solution that is going to cost you an hour of work for something that makes nobody lives easier. You're making this more difficult than necessary!

marcfehling:
The discussion moved away from the original topic of this PR towards the question how to maintain the CI of the publication list. Moving to github actions would enable us all to contribute to its maintenance and it would put work off @tjhei as in setting up the docker images. Of course, we could also leave it as is and follow the mantra to never change a running system to save our time.

marcfehling:
Build passes now.

PR #464, opened 2022-10-09
Add entry 2022:castelli.dorfler:comparison

gfcas:
Add our latest publication.

### DIFF
--- a/publications-2022.bib
+++ b/publications-2022.bib
@@ -192,6 +192,16 @@
   publisher = {Springer Science and Business Media {LLC}}
 }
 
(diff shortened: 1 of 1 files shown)


bangerth (approved):

marcfehling:
Thanks for adding this to the collection, and congrats on publishing. In the future, we would be grateful if you could also cite those papers that describe certain features of the library that you used in your work (see [here](https://www.dealii.org/publications.html#details)). In particular, we would appreciate references to the hp-papers, as well as in the future the matrix-free papers as you announced in the conclusions.

PR #466, opened 2022-10-10
Fix bibtool.rsc comment line

gfcas:
I am just playing around with bibtool for my own repo and found this.

### DIFF
--- a/bibtool.rsc
+++ b/bibtool.rsc
@@ -25,7 +25,7 @@ resource check_y
 % style improvements: no empty fields, fix page ranges, pure numerical fields
 resource improve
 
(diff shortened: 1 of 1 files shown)


marcfehling:
/rebuild

`bibtool` has many ways to declare comments, but thanks for figuring out this inconsistency.

marcfehling (approved):

gfcas:
> /rebuild
> 
> `bibtool` has many ways to declare comments, but thanks for figuring out this inconsistency.

I wasn't aware of the possible variations with bibtool, but I think it's pretty cool!

PR #479, opened 2022-11-03
Fix make sort

gfcas:
The current version in master deletes all the content of the ``publications-*.bib`` files on my machine. With the fix it works.

PR #482, opened 2022-11-11
Fix entry 2004:bangerth:framework

gfcas:
I am just trying things for #479 and found this. Maybe @bangerth knows best about this publication?!


marcfehling (approved):

bangerth:
I do. One of my very few solo-author papers :-)

PR #500, opened 2022-12-14
Fix field name.

marcfehling:
`article-number` is not a common field in bibtex. [doi2bib](https://www.doi2bib.org/bib/10.3390/geosciences12020050) reports this field as `pages`, so I changed it accordingly.

### DIFF
--- a/publications-2022.bib
+++ b/publications-2022.bib
@@ -966,7 +966,6 @@
   year    = 2022,
   volume  = 12,
   number  = 2,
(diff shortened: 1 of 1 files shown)


drwells:
This article has 17 pages (also numbered 1-17) - could we just delete the field instead? That seems less ambiguous.

drwells (approved):

drwells:
Thanks!

PR #503, opened 2023-01-19
Added three publications.

marcfehling:


publications-2023.bib
```diff
@@ -116,6 +130,16 @@ @Article{2023:paula.igreja.ea:numerical
   publisher = {Elsevier {BV}}
 }
 
+@Misc{2023:ulfsby.massing.ea:stabilized,
+  author  = {Ulfsby, Tale Bakken and Massing, André and Sticko, Simon},
```
bangerth, line 134: Let's see what happens to the accented letter here :-)

bangerth (approved):

publications-2023.bib
bangerth: Well, the CI doesn't like it...

marcfehling: oh, i missed that part. i guess the arxiv export doesn't translate unicode letters automatically.

PR #505, opened 2023-01-26
Update publications-2023.bib

JonasKnoch:
Added entry "2023:knoch:multi-scale"


marcfehling:
Thanks for reporting your publication! It is already part of the publication list
https://github.com/dealii/publication-list/blob/36ac84f237adb9587413c38586c0e719d9dda92d/publications-2023.bib#L104-L114

PR #512, opened 2023-03-08
Add 2023 publications to offline list.

marcfehling:
We forgot to add this line in #496.


tjhei (approved):

tjhei:
Not sure why we need this many passes, but I don't see any harm in increasing this value.

marcfehling:
> Not sure why we need this many passes, but I don't see any harm in increasing this value.

The number of passes were a bad call from my side. It turned out that there was an issue with one of the titles in the 2023 list, which got fixed in #513. Now, CI on the master branch passes.

PR #513, opened 2023-03-08
fix symbol

marcfehling:
Follow-up to #512.

The `\chi` symbol only works in math mode. We only spotted this after we included the 2023 publications into the global file. So my https://github.com/dealii/publication-list/pull/512#issuecomment-1460121117 was actually wrong.


marcfehling:
Looks like this now:
image

PR #517, opened 2023-05-31
Remove 9.3 from the list of suggested publications to cite

masterleinad:
Addressing https://github.com/dealii/dealii/pull/15290#discussion_r1212231629.

### DIFF
--- a/publications.include
+++ b/publications.include
@@ -57,56 +57,6 @@
       </li>
 
       <li>
-        Daniel Arndt,
-        Wolfgang Bangerth,
-        Bruno Blais,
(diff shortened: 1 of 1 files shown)


peterrum (approved):

marcfehling (changes requested):
Can you move the 9.3 article to the list of old release articles, i.e., here
https://github.com/dealii/publication-list/blob/72c4a8aee1fb46a3fba8cba463d44b0c0bd139c0/publications.include#L143-L145

marcfehling (approved):

PR #536, opened 2023-09-08
Update arxiv publications 2009 - 2017.

marcfehling:
Part of #289.

About a year ago, I started my mornings updating arxiv entries in our publication database to help me wake up. I stopped at the year 2018 for whatever reason, and I just rediscovered this stale branch.

So I spent a couple of minutes to rebase and sort out my commits to integrate my previous efforts to the database. I have a few other housekeeping commits from that time, which I would address in separate pull requests.

PR #537, opened 2023-09-08
Articles that mention, but do not use deal.II.

marcfehling:
Follow-up to #536. Part of #289.

I got suspicious when titles of articles clearly referred to other FEM libraries (as DUNE or FEMPAR), so I started to check articles for their content. I found a few that cite deal.II, but do not actually use it, and thus I suggest to remove them.

### DIFF
--- a/publications-2014.bib
+++ b/publications-2014.bib
@@ -531,11 +531,12 @@
   pages   = {108--126}
 }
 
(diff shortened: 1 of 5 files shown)


bangerth (approved):
This kind of change is of course unwelcome, but ok.

PR #538, opened 2023-09-08
Miscellaneous updates.

marcfehling:
Follow-up to https://github.com/dealii/publication-list/pull/536. Part of https://github.com/dealii/publication-list/issues/289.

Some various updates of entries that I stumbled over.

Found another thesis after looking up a preprint with the same name, which is nice.

### DIFF
--- a/publications-2015.bib
+++ b/publications-2015.bib
@@ -625,7 +625,16 @@
   title   = {Towards Accurate Modeling of Moving Contact Lines},
   journal = {arXiv:1510.06639},
   year    = 2015,
-  url     = {http://arxiv.org.ezproxy.library.tamu.edu/abs/1510.06639}
+  url     = {http://arxiv.org/abs/1510.06639}
+}
+
+@MastersThesis{2015:holmgren:towards,
+  author  = {Holmgren, Hanna},
+  title   = {Towards accurate modeling of moving contact lines},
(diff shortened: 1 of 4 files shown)


publications-2015.bib
```diff
@@ -625,7 +625,16 @@ @Article{2015:holmgren.kreiss:towards
   title   = {Towards Accurate Modeling of Moving Contact Lines},
   journal = {arXiv:1510.06639},
   year    = 2015,
-  url     = {http://arxiv.org.ezproxy.library.tamu.edu/abs/1510.06639}
+  url     = {http://arxiv.org/abs/1510.06639}
+}
+
+@MastersThesis{2015:holmgren:towards,
```
marcfehling, line 631: I am not sure whether a 'Licentiate thesis' in Sweden is equivalent to a PhD dissertation or Master's thesis. Do you have have an idea?

bangerth: Good enough. Not worth the time to worry about these sorts of things.

bangerth (approved):

publications-2015.bib
kronbichler: This is correct. For completeness, a licentiate thesis is half way between M.Sc. and Ph.D. (typically obtained after 2 years of Ph.D. studies and about 2 years before the defense).

PR #539, opened 2023-09-09
Removed three duplicates. Update preprint fields.

marcfehling:
Part of #289.

I found another stale branch that I haven't touched since a year or so. This mainly only contained information about preprint fields.

### DIFF
--- a/publications-2019.bib
+++ b/publications-2019.bib
@@ -1851,7 +1851,8 @@
   month   = jul,
   doi     = {10.1371/journal.pcbi.1007213},
   publisher = {Public Library of Science ({PLoS})},
-  editor  = {Qing Nie}
+  editor  = {Qing Nie},
+  preprint = {https://arxiv.org/abs/1901.00497}
 }
 
 @Article{2019:samii.kazhyken.ea:comparison,
--- a/publications-2019.bib
+++ b/publications-2019.bib
@@ -1851,7 +1851,6 @@
   month   = jul,
   doi     = {10.1371/journal.pcbi.1007213},
   publisher = {Public Library of Science ({PLoS})},
-  editor  = {Qing Nie},
   preprint = {https://arxiv.org/abs/1901.00497}
 }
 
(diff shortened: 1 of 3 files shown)


publications-2019.bib
```diff
@@ -1851,7 +1851,8 @@ @Article{2019:rudraraju.moulton.ea:computational
   month   = jul,
   doi     = {10.1371/journal.pcbi.1007213},
   publisher = {Public Library of Science ({PLoS})},
-  editor  = {Qing Nie}
+  editor  = {Qing Nie},
+  preprint = {https://arxiv.org/abs/1901.00497}
```
bangerth, line 1855: I don't think there is any use in listing the preprint. The *normative* publication is the one by the publisher; preprints may or may not coincide in text with the actual publication, and often are only early versions. I do not think that we should link to it.

Separately, the editor should also not be listed.

bangerth (commented):
It is possible that we had that discussion before -- in any case, I do not think that we should list preprints for published articles.

marcfehling:
We had the discussion in https://github.com/dealii/publication-list/pull/382#pullrequestreview-1093840819, which led to https://github.com/dealii/publication-list/pull/407.

You had the argument that preprints are no longer related to the final published manuscript, wheras I considered that preprints are showing the history of the manuscript and thus they are still related. Indeed, preprints are different from the final published manuscript, but the latter ones are in general hidden behind paywalls.

To allow users to quickly find a preprint version of the manuscript, we introduced the `preprint` field in #407. The `preprint` field contains a link to an old, but freely available manuscript. The `url` field still points to the published manusript, which is potentially hidden behind a paywall.

A different perspective: with the `preprint` field, we do not lose information about preprints by replacing them with the published manuscript, but we connect preprint and journal article together by keeping a link to the former. This also helps in maintenance of the publication list, as we can, e.g., still search for arxiv identfiers. In this PR, this helped me to find some preprint-duplicates, as the `preprint` field was already set for the published article.

I think it is fine to add the URL of the preprint, as long as we clearly flag that it is just a preprint. I believe that we do so sufficiently.

publications-2019.bib
marcfehling: > Separately, the editor should also not be listed.

For journals, I agree on this. I will remove the editor in this entry.

bangerth (approved):

PR #544, opened 2023-09-15
Update articles of Mohebujjaman.

marcfehling:
After going through some articles in the 2018 publication list, I spotted a couple of duplicates of Mohebujjaman, and many of the URL fields had links that lead to nowhere. I tried my best to update them.

### DIFF
--- a/publications-2022.bib
+++ b/publications-2022.bib
@@ -5,9 +5,9 @@
   title   = {Cavitation over solid surfaces: microbubble collapse, shock waves, and elastic response},
   journal = {Meccanica},
   year    = 2022,
-  month   = nov,
   volume  = 58,
   pages   = {1109--1119},
+  month   = nov,
   doi     = {10.1007/s11012-022-01606-5},
   url     = {https://link.springer.com/article/10.1007/s11012-022-01606-5},
   publisher = {Springer Science and Business Media {LLC}}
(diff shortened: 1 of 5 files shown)


publications-2022.bib
```diff
@@ -812,6 +812,21 @@ @PhDThesis{2022:mang:phase-field
...
+  year    = 2022,
+  volume  = 14,
+  number  = 5,
+  pages   = {1111--1137},
+  month   = jun,
+  doi     = {10.4208/aamm.oa-2021-0154},
+  url     = {https://global-sci.org/intro/article_detail/aamm/20554.html},
+  publisher = {Global Science Press},
+  preprint = {https://arxiv.org/abs/2101.06852}
+}
```
marcfehling, line 828: This article uses freefem, whereas an earlier version on researchgate uses deal.II. I will adjust it.

bangerth (approved):

bangerth:
@marcfehling I think I said this before, but I'll mention this again: This is not a good use of your time. Nobody really cares whether or not a publication from 2016 really does or does not use deal.II. Someone made the decision several years ago that it does, and it's not worth your or anyone else's time to go back and look at that decision again. If you want to help with the publication list, add recent publications and leave the ones from a few years ago alone -- it's not a productive use of your time.

PR #549, opened 2023-11-09
First run with the DOIs for 2022.

bangerth:


publications-2007.bib
```diff
@@ -229,6 +229,17 @@ @PhDThesis{2007:nigro:discontinuous
...
 }
 
+@Article{2007:nomura.sato.ea:multi-disciplinary,
+  author  = { Nomura, Tsuyoshi and Sato, Kazuo and Nishiwaki, Shinji and Yoshimura, Masataka },
+  title   = { Multi-disciplinary Multi-objective Topology Optimization of Electromagnetics and Structural Mechanics (For Case of Optimal Dielectric Resonator Antenna Designs) },
+  journal = { Trans. JSME, Ser.A },
+  year    = 2007,
+  volume  = 73,
+  pages   = { 1111--1119 },
+  issue   = 734,
```
marcfehling, line 239: `issue` is not BibTeX canon. Standard syntax is `number`. See section 4.2 Field in the documentation
https://mirrors.mit.edu/CTAN/biblio/bibtex/base/btxdoc.pdf

I believe the issue field only exists in biblatex, not bibtex.

As a consequence, the `issue` field will not be shown in `jabref`, see
https://github.com/dealii/publication-list/blob/c9fbd395b1c179fc0f75da6876da7d152fc67d5c/jabref-template/listrefs.layout

Yes, bibtex really *is* confusing.

publications-2007.bib
```diff
@@ -229,6 +229,17 @@ @PhDThesis{2007:nigro:discontinuous
   year    = 2007
 }
 
+@Article{2007:nomura.sato.ea:multi-disciplinary,
+  author  = { Nomura, Tsuyoshi and Sato, Kazuo and Nishiwaki, Shinji and Yoshimura, Masataka },
```
marcfehling, line 233: I'm not a fan of the spaces after and before the parenthesis. Primarily because I don't know how bibtex deals with them. Will they be ignored? Will they be treated as non-breaking spaces? I don't know

marcfehling (changes requested):
That's a good amount for a test run.

As mentioned privately, maybe we should try a different and (potentially) more reliable tool like https://github.com/bibcure/doi2bib or https://github.com/mseri/doi2bib for example.

marcfehling:
Actually, if you get the information from `api.crossref.org` instead of `doi.org`, you get the same results as doi2bib.

Let's talk about this tomorrow.

publications-2007.bib
bangerth: Let's not worry about this. This is how it comes out of the script, let's not spend too much work on these issues. If it renders poorly, we can always go back and change it again later, but I'm pretty sure that it will render the exact same way with or without the spaces.

bangerth:
I changed the issue->number thing. Let's not worry about the spaces issue.

marcfehling (approved):

bangerth:
@marcfehling Feel free to also merge :-)

marcfehling:
I thought I already did :)

PR #571, opened 2023-12-05
Add a format for @InCollection.

### LINKED ISSUE #397: broken formatting

bangerth:
Should fix #397.

### DIFF
--- a/jabref-template/listrefs.incollection.layout
+++ b/jabref-template/listrefs.incollection.layout
@@ -1,0 +1,107 @@
+\begingroup{year}
+  <tr class="show"><td><h3>\format[HTMLChars]{\year}</td></tr></h3>
+\endgroup{year}
+
+<tr id="\format{\bibtexkey}" class="entry">
+  <td>
+    \format[Authors(FirstFirst,Initials,Comma,Comma),HTMLChars]{\author}
+    <br>
+    <b>
+      \format[HTMLChars]{\title}
+    </b>
+
+    <br>
+
+   <!-- The following parts of each entry are all optional and may not
+        be present in all bibtex entries that are processed via the
+        current file. If present, the `\begin{field} ... \end{field}`
+        section will be expanded into something nonzero, and we will
+        do this with all of the following pieces of information about
+        each publication. Because we expect the first entry to always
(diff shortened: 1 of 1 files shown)


jabref-template/listrefs.incollection.layout
```diff
@@ -0,0 +1,106 @@
...
+        comment-end marker. 
+
+        Everything is closed with a period after the last entry.
+
+        -->
+
+       \begin{booktitle}In: \format[HTMLChars]{\booktitle}\end{booktitle}<!--
+       -->\begin{editor} (\format[HTMLChars]{\editor})\end{editor}<!--
+        -->\begin{pages}, pp. \format[FormatPagesForHTML]{\pages}\end{pages}<!--
+    -->\begin{publisher}, \format[HTMLChars]{\publisher}\end{publisher}<!--
```
marcfehling, line 34: I would also display the `note` field as in the other layouts.
```suggestion
        -->\begin{pages}, pp. \format[FormatPagesForHTML]{\pages}\end{pages}<!--
         -->\begin{note}  (\format[HTMLChars]{\note})\end{note}<!--
    -->\begin{publisher}, \format[HTMLChars]{\publisher}\end{publisher}<!--
```

marcfehling (approved):
Let's give it a try!

jabref-template/listrefs.incollection.layout
marcfehling: We can also debate whether the `volume` and `series` flag make sense, for example for this article:
https://github.com/dealii/publication-list/blob/f8e1b2d0c80361e1df64684bef278a936d01edc9/publications-2016.bib#L741-L752

Essentially, `InCollection` should be treated like `InProceedings`. I wonder whether `jabref` supports the use of one layout for multiple types.

bangerth: No, it does not. I copied this one from `@InProceedings`. Let's not make things more complicated than necessary. This is a step forward, it doesn't have to be perfect.

marcfehling (approved):

PR #576, opened 2024-01-16
Remove article that doesn't use deal.II

marcfehling:
In #575 I have added an article that actually does not use deal.II. Mea culpa.

They cited the parallel hp article and I guess I got too excited about that.

### DIFF
--- a/publications-2023.bib
+++ b/publications-2023.bib
@@ -811,21 +811,6 @@
   url     = {https://www.dzxb.org/en/article/doi/10.11939/jass.20210192}
 }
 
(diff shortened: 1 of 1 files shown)


marcfehling:
Anyways, we should keep that article in mind for the multigrid hp paper.

peterrum (approved):

bangerth:
Are you sure? Jake Harmon contributed a code gallery program on exactly this topic.

marcfehling:
They use FEM_2D which is written in RUST
https://github.com/jeremiah-corrado/fem_2d

PR #577, opened 2024-01-23
Add one more ASPECT publication.

bangerth:
Add one of the two publications from https://github.com/geodynamics/aspect/pull/5544 (the other one was already in the list).

### DIFF
--- a/publications-2023.bib
+++ b/publications-2023.bib
@@ -733,6 +733,17 @@
   publisher = {Copernicus GmbH}
 }
 
+@Article{2023:gea.mancilla.ea:overriding,
+  author  = {Gea, Pedro J. and Mancilla, Flor de Lis and Negredo, Ana M. and Hunen, Jeroen van},
+  title   = {Overriding plate thickness as a controlling factor for trench retreat rates in narrow subduction zones},
+  year    = 2023,
+  month   = mar,
+  note    = {Preprint},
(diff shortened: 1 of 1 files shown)


marcfehling (approved):
This is a preprint. Would you add a field `note = {Preprint}` to indicate it as such?

bangerth:
I'm not sure this is the right characterization. ESS Open Archive describes itself as follows:

> The Earth and Space Science Open Archive is a community server established to accelerate the open discovery and dissemination of earth, environmental, and space science research by archiving and sharing early research outputs, including preprints, presentations from major scientific meetings, and important documents of scholarly societies.

It might be that this is a preprint, but it's not obvious that it has ever been submitted anywhere else, nor whether this is an early version that may ever be superseded by a final version published elsewhere. My vote is to leave the entry as is, without adding any value judgment.

marcfehling:
If you scroll down on the [article website](https://essopenarchive.org/users/598340/articles/630961-overriding-plate-thickness-as-a-controlling-factor-for-trench-retreat-rates-in-narrow-subduction-zones), it states that it is indeed a preprint. According to your description, I assume there's a flag that the submitting author needs to set to mark it a preprint.
image

bangerth:
So done.

marcfehling (approved):

PR #588, opened 2024-06-06
Take over two from the ASPECT folks.

bangerth:
Taken from https://github.com/geodynamics/aspect/pull/5806. (That PR also updates one publication, but we already have that correctly in our database.)

### DIFF
--- a/publications-2024.bib
+++ b/publications-2024.bib
@@ -799,6 +799,21 @@
   publisher = {Elsevier BV}
 }
 
+@Article{2024:wiel.thieulot.ea:quantifying,
+  author  = {van der Wiel, Erik and Thieulot, Cedric and van Hinsbergen, Douwe JJ},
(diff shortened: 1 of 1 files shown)


publications-2024.bib
```diff
@@ -799,6 +799,21 @@ @Article{2024:vohra.peszynska:iteratively
   publisher = {Elsevier BV}
 }
 
+@Article{2024:wiel.thieulot.ea:quantifying,
+  author  = {van der Wiel, Erik and Thieulot, Cedric and van Hinsbergen, Douwe JJ},
+  title   = {Quantifying mantle mixing through configurational Entropy},
+  journal = {Solid Earth},
+  year    = 2024,
+  doi     = {10.31223/X56971}
+}
```
marcfehling, line 808: This is a preprint. Would you mark it as such with `note = {Preprint}`?

marcfehling (approved):

PR #592, opened 2024-06-10
Narasimhan 2604 my publication

narasimhan-2604:
I have now 
(1) Changed the months to have three characters only
(2) The urls have been removed from the dois
Thank you

### DIFF
--- a/publications-2023.bib
+++ b/publications-2023.bib
@@ -968,6 +968,20 @@
   publisher = {Springer Science and Business Media LLC}
 }
 
+@Article{2023:hussain.phanikumar.ea:mapping,
+  author  = {Hussain, Umair and Phanikumar, Gandham and Swaminathan, Narasimhan},
+  title   = {Mapping of multiphase field model parameters to physical factors in order to simulate desired phase transformations},
+  journal = {Computational Materials Science},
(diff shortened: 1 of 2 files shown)


marcfehling (approved):
The entries are okay now. The issue with the html check is unrelated to your patch. Thank you for the contribution!

marcfehling:
@tjhei The `Out of system resources` for the html test pops up more frequently, but it needs to pass so PRs can be merged. Can we remove the required tag from the html check in the meantime?

tjhei:
I will try to figure out what is going on.

marcfehling:
I have no idea why the CI keeps failing on this PR while other pass. Let me try something out.

masterleinad:
> I have no idea why the CI keeps failing on this PR while other pass. Let me try something out.

It says:
```
There is insufficient memory for the Java Runtime Environment to continue.
Cannot create GC thread. Out of system resources.
```

marcfehling:
I rebased the commits into one and ran the sorting script. I saw that one of the publications was already in our list, so this patch now contains two extra publications.

marcfehling:
CI passes now. Thanks for looking into it @tjhei

PR #612, opened 2024-11-26
Remove a file that's no longer used.

bangerth:
@tjhei says that `publications.include` is no longer used, though I have not been able to find how `script.sh` in @tjhei's docker image actually looks these days. Either way, if that's how it is, let's nuke the file to make sure we don't actually think that it needs to be kept up to date.


tjhei:
Agreed that we should clean this up.

But I am worried that this will break the CI workflow as the original .html used on the old website is still generated.

marcfehling (approved):

bangerth:
@tjhei Where does that happen? The file is not referenced anywhere in this repository, so it must be in a different repository. I suspect it's this line, right?
```
docker run --rm -v "$(PWD)/temp:/home/bob/source" tjhei/dealii-java-jabref
```
I just can't seem to find the script that this repo runs (`script.sh`).

tjhei:
I will take care of it in the next few days.

PR #650, opened 2025-09-29
use luatex and fix font handling

tjhei:
FYI @marcfehling

### DIFF
--- a/offline/publication_list.tex
+++ b/offline/publication_list.tex
@@ -10,14 +10,16 @@
 %
 
 \documentclass[]{article}
-\usepackage[utf8]{inputenc}
+\usepackage{iftex}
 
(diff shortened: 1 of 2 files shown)


blaisb (approved):

marcfehling (approved):
Looks good to me!

I tried to use `fontsetup` as a simple wrapper to plain `fontspec`, but its default font seems missing.
```
! Package fontspec Error: The font "NewCM10-Book" cannot be found.
```
Well it is just the font so I'm happy with whatever version works, since we only use it to verify our bibtex entries.

When trying `fontsetup` on my machine, it complained about a unicode `mu` to be missing in the font. I corrected an entry in the 2020 publication list to fix that warning.

PR #671, opened 2026-01-13
Add 11.

bangerth:
From the list at https://docs.google.com/document/d/1R9gFKys6_J1jeY9Yodws7k2FVgvA2AHVdEo59qMcxAA/edit?tab=t.0. I removed the ones I'm adding here from that list.

### DIFF
--- a/publications-2024.bib
+++ b/publications-2024.bib
@@ -41,6 +41,17 @@
   preprint = {https://arxiv.org/abs/2304.12032}
 }
 
+@Article{2024:almonacid.domnguez-rivera.ea:three-dimensional,
+  author  = { Almonacid, Javier A. and Dom\'{\i}nguez-Rivera, Sebasti\'an A. and Konno, Ryan N. and Nigam, Nilima and Ross, Stephanie A. and Tam, Cassidy and Wakeling, James M. },
+  title   = { A Three-Dimensional Model of Skeletal Muscle Tissues },
+  journal = { SIAM J. Appl. Math. },
+  year    = 2024,
+  volume  = 84,
+  number  = 3,
+  pages   = { S538--S566 },
+  doi     = {10.1137/22M1506985}
+}
+
 @Article{2024:alshehri.boffi.ea:unfitted,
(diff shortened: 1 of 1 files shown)


publications-2024.bib
```diff
@@ -356,6 +367,16 @@ @Article{2024:caldana.antonietti.ea:deep
...
+@Article{2024:caldana.antonietti.ea:deep*1,
+  author  = { Caldana, Matteo and Antonietti, Paola F. and Dede', Luca },
+  title   = { A deep learning algorithm to accelerate algebraic multigrid methods in finite element solvers of 3D elliptic PDEs },
+  journal = { Computers \&amp; Mathematics with Applications },
+  year    = 2024,
+  volume  = 167,
+  pages   = { 217--231 },
+  doi     = {10.1016/j.camwa.2024.05.013}
+}
+
```
marcfehling:
```suggestion
```
This is already in the list (see entry above). I must have missed that one during the initial screening. Mea culpa.

publications-2024.bib
```diff
@@ -41,6 +41,17 @@ @Article{2024:africa.fumagalli.ea:lifex-cfd
...
 }
 
+@Article{2024:almonacid.domnguez-rivera.ea:three-dimensional,
+  author  = { Almonacid, Javier A. and Dom\'{\i}nguez-Rivera, Sebasti\'an A. and Konno, Ryan N. and Nigam, Nilima and Ross, Stephanie A. and Tam, Cassidy and Wakeling, James M. },
+  title   = { A Three-Dimensional Model of Skeletal Muscle Tissues },
+  journal = { SIAM J. Appl. Math. },
+  year    = 2024,
+  volume  = 84,
+  pages   = { S538--S566 },
+  issue   = 3,
```
marcfehling:
```suggestion
  number  = 3,
```
Here and below: `issue` is not a valid bibtex field. Please use `number` instead.

bangerth:
So changed. I actually have a (different) doi2bib script that seems to work well and also translates accented letters correctly. Apparently it gets the issue vs number thing wrong.

PR #686, opened 2026-03-31
Add 3.

bangerth:
These all use a library called PHiLiP by @dougshidong which I believe uses deal.II.

@dougshidong -- are there other publications we're missing in our list?


bangerth:
@dougshidong Thank you very much for the feedback! If you still have influence on your advisor, would you suggest to them to cite deal.II in their papers? This way we can find them for the publications list.

Rombur (approved):

Issue #695, opened 2026-04-21
Add PHiLiP publications.


bangerth:
As pointed out by @dougshidong in #686, most of the papers listed at https://sites.google.com/view/mcgill-computational-aerogroup/publications actually use deal.II. We should add those that we don't already list, and keep revisiting that page just like we do for the ASPECT (#629) and DFT-FE (#620) publications.

bangerth:
Specifically, these are the 2022 ones mentioned in #686:

- [x] MacLean, K., & Nadarajah, S. (2022). [Anisotropic Mesh Generation and Adaptation for Quads using the Lp-CVT Method](https://www.sciencedirect.com/science/article/pii/S0021999122006404). Journal of Computational Physics, 111578.
- [x] Cicchino, A., Fernández, D. C. D. R., Nadarajah, S., Chan, J., & Carpenter, M. H. (2022). [Provably stable flux reconstruction high-order methods on curvilinear elements](https://www.sciencedirect.com/science/article/pii/S0021999122003217?casa_token=LfSfXj6eB70AAAAA:qch1VddkkJhrx5OLXxdKpX_14NnS-SekoyV9bsxHg7372unLbQsuY9aUzS6RoAt9VYAB3sSPygYa). Journal of Computational Physics, 111259.
- [x] Cicchino, Alexander, Siva Nadarajah, and David C. Del Rey Fernández. (2022) "[Nonlinearly stable flux reconstruction high-order methods in split form](https://www.sciencedirect.com/science/article/pii/S0021999122001565?casa_token=YLNWFsEB6Q4AAAAA:4_Ao0Y_87UhGCjPmBftB1wwXQbpNgRy8C_e-d4sKNorHerCxIPM90O5BPUakQ2I3BzDBDtwLYb8)." Journal of Computational Physics (2022): 111094.
- [x] Cicchino, A. & Nadarajah, S. "Nonlinearly Stable Split Forms for the Weight-Adjusted Flux Reconstruction High-Order Method: Curvilinear Numerical Validation." AIAA SCITECH 2022 Forum. 2022.
- [x] Thakur, P. & Nadarajah, S. (2022, June 5-June 8). Comparing Artificial Dissipation Based Shock Capturing Schemes for Discontinuous Galerkin Methods in the Presence of Curved Shocks [Presentation]. CSME 2022 International Congress, Edmonton, Alberta, Canada.
[...]

PR #702, opened 2026-04-22
Add the 2022 PHiLiP papers.

bangerth:
Not all of the things listed on the PHiLiP page seem to be publications. Some might just be presentations, or I could simply not find anything about these.

See #695.

### DIFF
--- a/publications-2021.bib
+++ b/publications-2021.bib
@@ -483,16 +483,6 @@
   url     = {https://onlinelibrary.wiley.com/doi/10.1002/nag.3167}
 }
 
-@TechReport{2021:cicchino.nadarajah:nonlinearly,
-  author  = {Alexander Cicchino and Sivakumaran Nadarajah},
-  title   = {Nonlinearly Stable Split Forms for the Weight-Adjusted Flux Reconstruction High-Order Method: Curvilinear Numerical Validation},
-  institution = {AIAA},
-  year    = 2021,
-  number  = {AIAA 2022-0190},
-  url     = {https://arc.aiaa.org/doi/10.2514/6.2022-0190},
-  doi     = {10.2514/6.2022-0190}
-}
-
 @Article{2021:clevenger.heister.ea:flexible,
   author  = {Thomas C. Clevenger and Timo Heister and Guido Kanschat and Martin Kronbichler},
   title   = {A Flexible, Parallel, Adaptive Geometric Multigrid Method for {FEM}},
--- a/publications-2021.bib
+++ b/publications-2021.bib
@@ -483,6 +483,16 @@
   url     = {https://onlinelibrary.wiley.com/doi/10.1002/nag.3167}
 }
 
+@TechReport{2021:cicchino.nadarajah:nonlinearly,
+  author  = {Alexander Cicchino and Sivakumaran Nadarajah},
+  title   = {Nonlinearly Stable Split Forms for the Weight-Adjusted Flux Reconstruction High-Order Method: Curvilinear Numerical Validation},
+  institution = {AIAA},
+  year    = 2021,
+  number  = {AIAA 2022-0190},
+  url     = {https://arc.aiaa.org/doi/10.2514/6.2022-0190},
+  doi     = {10.2514/6.2022-0190}
+}
+
 @Article{2021:clevenger.heister.ea:flexible,
   author  = {Thomas C. Clevenger and Timo Heister and Guido Kanschat and Martin Kronbichler},
   title   = {A Flexible, Parallel, Adaptive Geometric Multigrid Method for {FEM}},
--- a/publications-2022.bib
+++ b/publications-2022.bib
@@ -1887,13 +1887,6 @@
   doi     = {10.5220/0011135200003274}
 }
 
-@InProceedings{2022:pethrick.nadarajah:fully,
-  author  = {Pethrick, C. and Nadarajah, S.},
-  title   = {Fully Discrete Explicit Entropy-Stable Flux Reconstructions Schemes for Nonlinear Scalar Conservation Laws},
-  booktitle = {CSME 2022 International Congress, Edmonton, Canada},
-  year    = 2022
-}
-
 @PhDThesis{2022:petrasova:pokrocile,
   author  = {Petr\'a\v{s}ov\'a, Iveta},
   title   = {Pokro\v{c}il\'e metody anal\'yzy a optimalizace matematick\'ych model\r{u}},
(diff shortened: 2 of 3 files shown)


bangerth:
I also added the remaining AIAA papers.

publications-2023.bib
```diff
@@ -514,6 +524,26 @@ @Misc{2023:bohn.dorfler.ea:adaptive
...
+@TechReport{2023:brillon.nadarajah:on*1,
+  author  = {Julien Brillon and Sivakumaran Nadarajah},
+  title   = {On the Use of Nonlinearly Stable Flux Reconstruction for Implicitly Filtered Large Eddy Simulation},
+  institution = {AIAA},
+  year    = 2023,
+  number  = {AIAA 2023-1224},
+  url     = {https://arc.aiaa.org/doi/10.2514/6.2023-1224},
+  doi     = {10.2514/6.2023-1224}
+}
+
```
marcfehling:
```suggestion
```

Duplicate of the article above.

publications-2022.bib
```diff
@@ -1877,6 +1887,13 @@ @InProceedings{2022:peters.besseron.ea:innovative
   doi     = {10.5220/0011135200003274}
 }
 
+@InProceedings{2022:pethrick.nadarajah:fully,
+  author  = {Pethrick, C. and Nadarajah, S.},
+  title   = {Fully Discrete Explicit Entropy-Stable Flux Reconstructions Schemes for Nonlinear Scalar Conservation Laws},
+  booktitle = {CSME 2022 International Congress, Edmonton, Canada},
+  year    = 2022
+}
```
marcfehling: Please add a note that this is an abstract and add the DOI https://doi.org/10.7939/r3-e8na-k444

publications-2022.bib
```diff
@@ -2234,6 +2251,13 @@ @PhDThesis{2022:stolte:-mera
   url     = {https://hdl.handle.net/1969.1/197758}
 }
 
+@InProceedings{2022:thakur.nadarajah:comparing,
+  author  = {Pranshul Thakur and Siva Nadarajah},
+  title   = {Comparing Artificial Dissipation based Shock Capturing Schemes for Discontinuous Galerkin Methods in the presence of Curved Shocks},
+  booktitle = {CSME 2022 International Congress, Edmonton, Canada},
+  year    = 2022
+}
```
marcfehling: Please add a note that this is an abstract and add the DOI https://doi.org/10.7939/r3-3497-kz98

marcfehling (changes requested):
You have added a few entries as tech reports. From my understanding, those are conference proceedings, which is also reflected in the doi metainformation. I would suggest to treat them this way.

publications-2022.bib
marcfehling: The abstract does not mention neither deal.II nor Philip.

publications-2022.bib
marcfehling: Same here. The abstract does not mention neither deal.II nor Philip.

bangerth: I don't have access to the paper, and simply went with @dougshidong's comment that all but the ones he specifically listed use PHiLiP.

Either way, since it's just an abstract, perhaps I just remove it.

publications-2022.bib
bangerth: I removed this one as well.

bangerth:
I fixed the three wrong entries, but on the issue of tech report vs proceedings: I still think that the former is the right classification. These are technical reports, individually numbered, produced as an annual series by a professional organization. It is *also* true that they are the result of a conference. So perhaps the issue is debatable, but it is not worth your or my time to figure out what the *perfect* solution on this is -- we provide a way for people to find these articles, and I think that's good enough :-)

bangerth:
Ping?

marcfehling:
> on the issue of tech report vs proceedings: [...]

I typically just go with the metainformation provided by DOI registration agencies. That is the response we get for example with our own [doi2bib.py](https://github.com/dealii/publication-list/blob/master/scripts/doi2bib.py) script or [doi2bib.org](https://www.doi2bib.org/) through [DOI content negotiation](https://citation.doi.org/docs.html).

These agencies classified the AIAA articles as proceedings articles. The bibtex entries we get through content negotiation lack the `number` field, which is, however, included in the DOI as the last 8 digits.

I do not feel strongly about the classification, and I am fine with keeping them as technical reports. I just want entries to have a persistent reference (doi/handle), which they all provide in this patch.

marcfehling (approved):

publications-2021.bib
```diff
@@ -483,6 +483,16 @@ @Article{2021:choo.semnani.ea:anisotropic
...
 
+@TechReport{2021:cicchino.nadarajah:nonlinearly,
+  author  = {Alexander Cicchino and Sivakumaran Nadarajah},
+  title   = {Nonlinearly Stable Split Forms for the Weight-Adjusted Flux Reconstruction High-Order Method: Curvilinear Numerical Validation},
+  institution = {AIAA},
+  year    = 2021,
+  number  = {AIAA 2022-0190},
+  url     = {https://arc.aiaa.org/doi/10.2514/6.2022-0190},
+  doi     = {10.2514/6.2022-0190}
+}
```
marcfehling: This was published online in 2021, but the Forum happened in 2022 and the articles number also reflects an association with 2022. It is also listed as a 2022 publication in #695.

Do you want to adjust the year field and move it to the 2022 list?

bangerth: I don't know. It says "Published Online:29 Dec 2021", so I thought 2021 is the right choice. But then the copyright on the page says 2022.

In the end, I feel the same way as about much else on this topic: I want to provide a list of resources for people to find papers that show results obtained with the help of deal.II, and to provide ourselves with evidence how widely it is used. I do not want to spend time arguing about minutiae -- that is not time well spent because I do not think it actually makes anything better.

bangerth: Let me not argue. I move the article to 2022.

blaisb: Peace and love ❤️

blaisb (approved):

marcfehling (approved):

publications-2021.bib
marcfehling: I did not mean to be a source of frustration.

marcfehling:
I believe Jenkins CI is stuck.

blaisb:
> I believe Jenkins CI is stuck.

Yeah you are right....
Is there a way we can restart it or we just have to wait?

bangerth:
It seems to have recovered.

PR #706, opened 2026-05-20
Add 3.

bangerth:


publications-2026.bib
```diff
@@ -359,6 +372,17 @@ @Article{2026:hochbruck.scheifinger:local
...
+@Article{2026:huang.wang:general,
+  author  = { Huang, Weizhang and Wang, Zhuoran },
+  title   = { A general regularization strategy and convergence analysis for the iterative solution of singular Stokes problems },
+  journal = { Communications in Nonlinear Science and Numerical Simulation },
+  year    = 2026,
+  volume  = 161,
+  pages   = 110059,
+  doi     = {10.1016/j.cnsns.2026.110059},
+  url     = {http://doi.org/10.1016/j.cnsns.2026.110059}
+}
```
marcfehling: They just use the manufactured example from step-56 instead of deal.II itself. For their computations they use Matlab.

Does this count? In the past I have omitted articles that were only adopting the scenarios presented in the examples. I'm not opposed to adding this; I just want to make sure we filter articles with the same criteria.

marcfehling (approved):
Looks good to me!

publications-2026.bib
bangerth: I couldn't quite figure that out. The say that they use a test case "based on step-56". I took that as perhaps saying that they used the matrix from there since Zhuoran Wang uses deal.II. But you may be right. I removed the publication.

marcfehling (approved):

publications-2026.bib
bangerth: But actually, we could also ask her herself: @sophy1029?

PR #715, opened 2026-08-11
Update the DFT-FE publications.

bangerth:
This should get us past 3000.

### DIFF
--- a/publications-2025.bib
+++ b/publications-2025.bib
@@ -130,6 +130,18 @@
   copyright = {Creative Commons Attribution 4.0 International}
 }
 
+@Article{2025:baek.das.ea:quasicrystal,
+  author  = { Baek, Woohyeon and Das, Sambit and Tan, Shibo and Gavini, Vikram and Sun, Wenhao },
+  title   = { Quasicrystal stability and nucleation kinetics from density functional theory },
+  journal = { Nat. Phys. },
+  year    = 2025,
+  volume  = 21,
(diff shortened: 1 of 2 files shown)


bangerth:
Ping?

bangerth:
Ping again?

marcfehling (commented):
Looks good to me. I just have one suggestion: `issue` is not a valid bibtex field. Please use `number` instead.

bangerth:
I fixed those. It's a bummer that doi2bib (the version on my system) keeps generating these invalid entries :-(

bangerth:
Perhaps we could add a check for these invalid fields to the CI. This way the submitter can fix this and you don't have to look out for it with every patch.

marcfehling (approved):

marcfehling:
> It's a bummer that doi2bib (the version on my system) keeps generating these invalid entries.

IIRC `issue` is a field that was introduced with biblatex. biblatex is compatible with bibtex, but not the other way round. So using `number` is the right choice for our publication list IMO. Could it be that your personal doi2bib script is tuned for biblatex?

> Perhaps we could add a check for these invalid fields to the CI.

Certainly. Alternatively we can also instruct bibtool to rename these fields with `rename.field{issue=number}`. See also section A.12.4 of the [bibtool docs](https://gerd-neugebauer.de/software/TeX/BibTool/bibtool.pdf).

PR #716, opened 2026-08-12
Add a script that creates a graph with the number of publications per year.

bangerth:
100% vibe coded. Remarkably easy. Output looks like this: 
publications_per_year

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,17 @@ Library path: .:/path/to/bibtool/lib/BibTool
 Special configuration options: none
 ```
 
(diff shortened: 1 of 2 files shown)


bangerth:
I pushed a second commit that marks incomplete years with gray bars, like we do for the webpage.

bangerth:
Ping?

marcfehling (approved):
Looks good to me. May I suggest to move the python script to the `scripts` folder.

bangerth:
So moved.

